PR #51, opened 2023-03-06
Feature/user avatars

KennethMetz:
Adds avatar and bio section to /profile.


KennethMetz:
- Let me know if the UX to edit the avatar/bio work for you then I'll fine tune the <TextField> styling and work on adding more avatar options.

- Also, is there a size limit we want to shoot for on each of these avatar images?  (I've been trying to keep them around 20kb, but 🤷 )

metzsg (approved):
This looks great, I like the UX of the avatar and bio editing a lot.  Very clean.  Looking forward to building on this momentum with more profile page features/improvements.

metzsg:
Regarding avatar size, since we render them 80x80 as the largest size in the mocks right now, it would make sense to make all our avatar images 160x160 (for retina displays) and then save them all as jpgs with reasonably high quality.  I was able to get images around 12kb and still looking _expensive_ on the site.  Image size will vary depending on image content somewhat, but I think it should be fine to start here and see how things go.  I'll follow up with some avatar conversions in another PR.

PR #53, opened 2023-03-08
Adds horizontally scrolling shelves to avatar selection menu.

KennethMetz:
This is an attempt to organize the avatar selection menu and make it feel BIG.

- As we add more shelves, I would like to eventually include a vertical "see more" button as well.


metzsg (approved):
Looks great!  Love the new avatars, the sets really look good together

PR #56, opened 2023-03-13
Fix bug in updating analysis, add live updates

metzsg:
This commit fixes a bug where changes to the user's likes or dislikes while on DetailedView would not be reflected in future analysis, and also adds the ability to reload analysis live on changes.  I've added a cool little loading indicator when you switch anime, using BreathingLogo, to make it clear when the analysis is outdated and pending updates.   The animations in score bars when the scores switch between anime is kinda cool in that it lets you compare them based on how much the bars move.


KennethMetz (approved):
Siiiiiiick!!! That animation looks great!

I've been trying to think of a way to trigger the animation on cached pages too but have struck out so far.  Any suggestions?

PR #57, opened 2023-03-15
Format header to have tabs

KennethMetz:
This commit adds tab navigation to the header and hides the search bar behind a toggle.
- There is a mobile use case where guest accounts will have two "user" icons shown on the header (the profile tab and the dropMenu).  I'm thinking of adding a different default guest image for the dropMenu...maybe a "G" or a question mark...


metzsg (approved):
Nice! This is looking great.  I think this really improves the discoverability of the navigation.

I agree that it would be good to find something to avoid the double User icons, worth thinking about more

PR #58, opened 2023-03-16
Header tab styling

metzsg:
I was super inspired by these tabs!  This pr just tweaks the tabs slightly for various styling purposes:
- Icons on mobile are now text.primary
- Tabs are a bit more centers in large widths.
- Refactors tab code into a HeaderTab component to clean up Header a bit and enable 'active tab' styling more easily.
- Added an underline to the active tab.  Still messing with this one, not 100% sure how the active tab could be differentiated, but this is one idea.


KennethMetz (approved):
Good stuff man! Also clever use of :after to create the active tab underline!

PR #59, opened 2023-03-16
Add Top 8 List

KennethMetz:
This commit adds an interactive top 8 list.
- Styling of profile to match the latest mock will be completed in a follow-up commit.


metzsg (approved):
THIS IS AMAZING!  Love the way this turned out, the drag-and-drop is great ux for managing the list.  Code looks good, love the refactoring here to break out complex parts of the Profile page into their own components.  Let's go!

PR #62, opened 2023-03-20
Adds swipe functionality for shelves on home page

KennethMetz:
- This is my stab at adding swiping functionality to our anime shelves.  I know you mentioned you were looking into this...so if I missed something or am stepping on your work at all feel free to trash this pr.


metzsg (approved):
This is definitely an improvement, so I like it.  Eventually, I'm hoping to get the shelves to actually have content that scrolls as you swipe, but this is a good solution until then.

PR #69, opened 2023-03-26
Redesign profile page, add Lists detail pages

metzsg:
Some notes:
- Two new dependencies added (`slug` and `material-ui-confirm`), so be sure to run `npm install` while on this branch.
- Suggested items are shown beneath watchlists, which is cool.  I don't show suggested items beneath Likes/Dislikes lists right now, since suggestions based on Likes is basically the same as For You and showing the user more items they might dislike seemed like the opposite of the point of the site.  Maybe there is a way it could be funny, or help people refine their preferences, so we can consider this more.
- The avatar selection and top8 are now squeezed into the sidebar on wide screens. I don't love it. I made the top8 text size smaller just so titles had a better chance of fitting. I may mess around with returning to a single-column look in a follow-up commit to allow these to breathe more.
- The list detail pages on mobile are kinda not great UX since you navigate there and then have to scroll a lot to see the actual list (past the avatar, bio, top8). I also don't love this. This also makes me consider simplifying the list detail pages to make them work better on mobile.
- I've added a confirmation to deleting watchlists, since this can be a pretty destructive action.
- I've added list name slugification, but I do want to do more in the future to avoid slug collisions if the user has two lists with the same name, etc.
- I haven't yet implemented the ability to view other profiles, and therefore I don't use the full `/profile/<userId>` paths yet. That is coming next.
- I have a few more features that I think would be awesome: drag-and-drop to reorder watchlists, dragging items from watchlists or likes into the top8, and other stuff. Expect more commits soon as I iterate on these ideas!


KennethMetz (approved):
Niiice!!!  Awesome stuff!

And good catch on the slug collisions...I guess my form validation function checking for duplicate names is case sensitive.  But, feel free to tinker with the Yup schema if you think it'd be easier to address there...

PR #70, opened 2023-03-30
Set up system of avatar 'names' to store in LocalUser

metzsg:
Webpack has been configured to take all image names and append a hash to the end of them based on the image file contents (https://drive.google.com/file/d/1BLYsi8Xc9tbgNBaIjM7IIMJZDIZG0CCV/view?usp=sharing).  For example, `sakura.jpg` became `sakura.e3daff7eca305410d3c4.jpg`.  This hash will be stable, but will change if the image is changed, which busts browser caches since the browser will not have cached the new filename, so browsers will immediately download the new version.

Unfortunately, this means that if we store the filename in the `avatar` field of LocalUser, then we update the image for one of our avatars, their avatar will break.

I've solved for this by creating a simple 'name' string for each avatar, then creating a map of name -> url in a new file called `Avatar.js`.  That file exports a function called `getAvatarSrc(name)` which will convert each simple name into the full image path needed to display it in a given build.  This means you won't be able to do `src={localUser.avatar}` anymore, but instead, `src={getAvatarSrc(localUser.avatar)}`.

This is a *breaking change* that will make all existing users' avatars break until they update them.  But after they do, the avatars should never break again.  Unless we do something dumb like change remove an avatar or change its name.

### DIFF
--- a/src/Components/AvatarIcon.js
+++ b/src/Components/AvatarIcon.js
@@ -1,9 +1,10 @@
 import { Avatar, IconButton, useTheme } from "@mui/material";
-import { useContext } from "react";
+import { useContext, useMemo } from "react";
 import { useAuthState } from "react-firebase-hooks/auth";
 import { LocalUserContext } from "./LocalUserContext";
 import { auth } from "./Firebase";
 import { SaveToFirestore } from "./Firestore";
+import { getAvatarSrc } from "./Avatars";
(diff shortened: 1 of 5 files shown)


KennethMetz:
This works for me - great foresight spotting how this was going to paint us into a corner!!

PR #77, opened 2023-04-04
Adds 'see more' button for reviews

KennethMetz:
- Detailed View page now displays (4) reviews on page load, and another (4) with each 'see more' click.
- I would like to add the ability to filter reviews in an upcoming commit.

### DIFF
--- a/src/Components/DetailedView.js
+++ b/src/Components/DetailedView.js
@@ -353,7 +353,7 @@ export default function DetailedView() {
             style={{ ...subheadStyle, display: "flex", alignItems: "baseline" }}
           >
             Reviews{" "}
-            {animeReviews.length > 2 ? (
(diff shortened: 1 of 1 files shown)


metzsg (approved):
Looks great! I like the pagination of reviews, it is a good pattern and will help the site grow.

I added a null-check on some uses of `animeReviews`, to avoid some runtime errors I hit.

As a future improvement, it might also be worth thinking about a way to paginate the reviews as they are obtained from the server, only loading one page at a time.  I imagine a case where there are 200 reviews for a show.  Reading the whole set on page load would be slow, and would also charge us for 200 reads on each page load, when the user may only want to see a handful of reviews.  Check out https://firebase.google.com/docs/firestore/query-data/query-cursors#paginate_a_query

PR #78, opened 2023-04-04
Feature/other profiles

metzsg:
This PR:
- Simplifies the hooks to get a profile from the API.
- Switches to use the `/profile/<userId>` path structure
- Adds links to user profiles via the avatars of review authors
- Introduces a new `ProfilePageContext` concept to abstract out logic around when the profile page controls should be editable, and what data to show (we use LocalUser for the user's own profile, and the fetched profile for others, but to program a profile page UI, you don't have to worry about that while using this context)

The last part is interesting, in specific. Since profile pages can now be rendered from localUser or fetched public profile data, and I didn't want to duplicate components to display these, I have abstracted all logic of getting profile data, determining whether the user can edit the profile, and applying these edits, to a context called ProfilePageContext. This means individual profile page components should not obtain localUser or write to localUser, but instead do all interaction with LocalUser through the ProfilePageContext. I'm migrated all components on the Profile page to use this context, except for the avatar selection, which still does some manual LocalUser manipulation when you choose avatars.

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -99,13 +99,8 @@ export function useGetTitles(url) {
   );
 }
 
-export function useFetchProfile(userID) {
-  return useGetProfile(
-    `https://api-jet-lfoguxrv7q-uw.a.run.app/profile/${userID}`
-  );
-}
-
-export function useGetProfile(url) {
+export function useProfile(userID) {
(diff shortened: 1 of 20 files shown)


KennethMetz (approved):
Awesome feature!!!

Nice use of `ProfilePageContext` for both the users own profile and for others. Really slick stuff.

PR #79, opened 2023-04-06
Filtered Reviews

KennethMetz:
ASK AND YE SHALL RECEIVE (I am expecting **HEAVY** site traffic once we launch, so let's minimize these Firestore reads!!)

Two updates, based on your comments:
1. Only reads (4) documents from Firestore at a time.
2. Adds a button to filter reviews.

Note:
- I couldn't figure out if it's worth caching these Firestore calls with a TanStack Query.  It seems unlikely someone would tinker with the filters enough to burn a significant number of reads....but I'm sure stranger things have happened...

### DIFF
--- a/src/Components/DetailedView.js
+++ b/src/Components/DetailedView.js
@@ -28,7 +28,6 @@ import { auth } from "./Firebase";
 import {
   GetPaginatedReviewsFromFirestore,
   PopulateFromFirestore,
-  PopulateReviewsFromFirestore,
 } from "./Firestore";
 import LikeButtons from "./LikeButtons";
 import { LocalUserContext } from "./LocalUserContext";
@@ -72,8 +71,6 @@ export default function DetailedView() {
   }, [user, anime, loading]);
 
   useEffect(() => {
-    console.log(sortOption);
-
     GetPaginatedReviewsFromFirestore(
       anime,
       animeReviews,
@@ -81,6 +78,7 @@ export default function DetailedView() {
       sortOption,
       lastVisible,
       setLastVisible,
+      seeMore,
       setSeeMore
     );
   }, [location.pathname, anime, sortOption]);
@@ -375,7 +373,7 @@ export default function DetailedView() {
               }}
             >
               Reviews{" "}
-              {animeReviews?.length > 2 ? (
+              {animeReviews?.length > 2 && (
                 <Typography
                   sx={{
                     fontFamily: "interMedium",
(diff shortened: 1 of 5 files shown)


metzsg (approved):
Nice! This looks great.  I absolutely feel like DetailedView pages are going to be the most-loaded kind of page by any user of the site, so making it cheaper and faster to load is a real win.  And I love the order-editing feature, it is going to be really useful.  I don't think we need to worry about caching these yet.  Especially now that we are loading only 4 documents per page, this should be pretty efficient.

There is one small bug that I found, that has to do with how React works with component re-use.  Basically, when navigating from one DetailedView page to another, React just re-uses the same `DetailedView` component, and doesn't make a new one or reset the states.  This means that, if the last DetailedView has more content than can fit on one page, the `lastVisible` field will be non-null when `GetPaginatedReviewsFromFirestore` gets called for the new anime, and `animeReviews` won't get cleared out and the new reviews will be added to the end of the reviews from the last anime page.  So you see reviews for a different show on other anime pages.  Navigating from FMA Brotherhood to any other show then scrolling down to the reviews will demonstrate this.

I can think of two solutions.
1) Do some workaround in DetailedView.  Probably just keep track of the animeId that `animeReviews` is for, and when you are parsing the new page of reviews, if the animeId these new reviews were requested for doesn't match the one from before, don't copy over the old AnimeReviews.

2) Prevent DetailedView re-use silliness by wrapping DetailedView in a second component, something like `DetailedViewPage`, and that component will basically be something like:
```
  const params = useParams();
  const animeId = params.animeId;
  return <DetailedView key={animeId} />;
```
This will cause the key to change whenever the user navigates to a new anime page, and so a new DetailedView will be instantiated and we don't have to worry about this surprising component re-use.  This might make the component easier to work with and reason about in the future, but we do lose the cool animation of the score bars when you get to the new anime page.  We could always rig it up to animate the bars up from 0 on each DetailedView instead, which could still be cool and flashy.  I have this illustrated in a commit in branch [feature/paginated-review-queries-with-fix](https://github.com/KennethMetz/anime_finder/tree/feature/paginated-review-queries-with-fix).

I'll leave it up to you which of these you want to use.  I think I might have a slight preference for 2 but we could always do that later too.

KennethMetz:
Great catch!!  Also utilizing a changing `key` property is a really great fix! It's waaay less "hacky" than what I was first picturing.

Also, I really like the idea of each score bar filling up from 0 on each page load - the more movement the better in my book!

PR #80, opened 2023-04-06
Make review content grid grow to fill available width

metzsg:
Otherwise, when the review text is really short, the review content ends up being small and centered in the review card (the super short review coolman1 has on FMA Brotherhood will demonstrate this).  This change just makes the main grid of the review expand to fill available width.


metzsg:
Responsive reviews look great, btw!

PR #83, opened 2023-04-07
Animate ScoreBars to their initial positions

metzsg:
Since the LinearProgress components only animate _changes_ in value, once the scores are loaded, I start by multiplying all their values by a multiplier of 0, then change the multiplier to 1 after a short 100ms delay.  This causes a change in value, so the bars animate from 0 to the full score values.


KennethMetz:
Very nice - I'm a big fan of this feature!!!

PR #84, opened 2023-04-07
Centers like buttons on their shelf.

KennethMetz:
This keeps them from breaking into (2) rows when the anime tiles get very small.  
This would occur right before the 600px {sm} breakpoint.

### DIFF
--- a/src/Components/DetailedView.js
+++ b/src/Components/DetailedView.js
@@ -247,7 +247,8 @@ export default function DetailedView() {
               xs={12}
               md={3.5}
               sx={{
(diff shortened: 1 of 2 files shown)


metzsg (approved):
LGTM!  I agree that the AnimeCard hover content gets unpleasant when the cards are small.  This is a big improvement.  The watchlist pages are still the big offender of having small cards since the shelf doesn't have the full width but the number of cards is sort of preset to work when the shelf goes full width -- I have an idea to improve this in a follow-up.

PR #85, opened 2023-04-07
Clean up Avatar display

KennethMetz:
-Changes guest avatar to a "G" instead of the phosphor `User` icon.  This resolves having (2) User icons in the header for the compressed mobile view.

-Shows the users avatar in the Drop Menu's `ListItemButton` (the first item in the drop menu - now it shows the avatar instead of the first letter of the user's name)


metzsg (approved):
LGTM!  This one is nice, too.  I like using the initial for guests and users without a selected avatar.

PR #86, opened 2023-04-07
Firefox styling clean-up

KennethMetz:
Fixes (2) display issues with the app in Firefox:
1. Removes styling from all links.
2. Centers typography in Dropmenu Avatar (when it fails loading a link and displays an initial instead...ie, for a "guest" account).


metzsg (approved):
Good catches.  I have only tested the web app so far in chrome and safari, but it would be good to regularly check it in firefox and edge as well

PR #88, opened 2023-04-08
Update useAnimeMPTW to use planned_to_watch sort

metzsg:
Not sure if this is a bug or if it is intended, but I was looking through APICalls.js and I realized that `useAnimeMPTW` uses the `most_rated` sort order, although tI think he title indicates that it was supposed to be `most_planned_to_watch`.  The results are shown on home under the 'Most Buzzed About' header.  If this is working as intended, then I'll just delete this branch, but otherwise, correcting this will add a little more diversity to Home.


KennethMetz:
Great catch!! I think that's a typo from when I changed the API calls over to use TanStack Query.  
The added diversity is a nice change!

PR #90, opened 2023-04-08
Adjust layout of profile pages to fix some issues

metzsg:
I am still messing around with this new layout but let me know what you think.  Major changes and motivations:

- The list pages now don't have a sidebar.  This was done to work around the issue where our anime shelf ends up with really small cards when it doesn't have the full page width to use.  This makes the text and buttons on the cards overflow and makes them harder to use.  Check out the 'suggested' shelf on watchlist pages on the previous designs to see what I mean.
- The sidebar on the main profile page is now wider.  This means that the top 8 list can have larger text and be more prominent and feel less squished.
- The user banner is now a page header that spans the full width.  This should behave better with the avatar editor, which now has the full screen width to use (we might even want to constrain it to a max width that is narrower), as well as long user display names are much less likely to truncate.  We can put action buttons to the right later, if we have profile-level interactions we want to include.  I can't think of any, but there is a bunch of space there now.
- The user display name in the profile user banner is now "page header" sized, which matches the anime titles on the DetailedView pages and the "For You" on home, creating a more consistent design language of a page title being 40px, extra-bold.

Not sure if it necessarily looks _better_ than the previous layout, but I think it does solve some UX issues, especially on small screens.  Check it out and lmk what you think.


KennethMetz:
Very nice - I really like the re-work!! And great solution for those tiny recommended cards!

One thought I had:
- On the main profile view, I'm not totally happy with the look of the "Bio" in the sidebar. Not having a background makes it look/feel a bit off to me. Maybe we could try either putting the Bio on it's own gradient card or on the Top8 gradient card (if we do this we could also print other user info on it as we go...like the # of reviews, badges, etc...)?

KennethMetz (approved):

PR #91, opened 2023-04-10
Responsive Filter Button

KennethMetz:
Makes the 'filter by' button (in the reviews section) shrink down on small devices.


metzsg (approved):
Let's gooo!!!

PR #92, opened 2023-04-10
Adds a shelf of Pokemon avatars

KennethMetz:


metzsg (approved):
LGTM! I also like the change that shows all additional avatars when the user clicks 'see more' instead of breaking it into multiple pages

PR #94, opened 2023-04-11
Remove top-level MUI imports

KennethMetz:
Replaces all top-level imports of MUI with path imports to avoid pulling in unused modules and reduce build size.

`// 🚀 Fast`
`import Button from '@mui/material/Button';`
`import TextField from '@mui/material/TextField';`

`// 😭 Slow;`
`import { Button, TextField } from '@mui/material';`


metzsg (approved):
Nice!!  This will definitely help reduce our bundle size.

PR #95, opened 2023-04-11
Replaces .gif with .mp4 for faster page load

KennethMetz:
On /landing page replace our .gif GIF with a .mp4 to speed up the initial render.


metzsg (approved):
Great idea, this is a huge size savings while not reducing quality.  We should do this for all animated images if possible!

PR #97, opened 2023-04-11
Corrects guest name to be capitalized

KennethMetz:
- Corrects the 'Guest' user.name to be capitalized when writing to firestore, so it will appear correctly when displayed on /profile


metzsg (approved):
Looks good!  We can also try to fix the “weird stuff flashes while the profile data loads” by adding in a “ghost” version of these user banners when isLoading is true from the profile page context.  I didn’t add it just cause there were now two user banners after I added the small one and I wanted to think about if there was a cleaner way.  Maybe this is it 😀

PR #99, opened 2023-04-12
Replaces .ttf fonts with .woff files

KennethMetz:
Note: I couldn't figure out how to get the variable fonts to work...but we could get even more space savings if we figure that out.  So, I'll circle back to this PR.


metzsg (approved):
Looks good!  Great size reduction

PR #100, opened 2023-04-12
Adds fallback fonts

KennethMetz:
- Had some trouble implementing this in the ways you recommended...but this seems to still do the trick.  Lmk if that's not the case...

### DIFF
--- a/src/Components/theme.js
+++ b/src/Components/theme.js
@@ -112,5 +112,8 @@ export function createAppTheme(darkMode) {
         ],
       },
     },
+    typography: {
+      fontFamily: ["Helvetica", "Arial", "sans-serif"],
+    },
   });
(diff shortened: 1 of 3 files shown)


metzsg (commented):
This commit is a good step in the right direction, since it cleans up some not-desired font stuff and starts setting up some typography in the theme.  I do still see the default serif font during load, although for much shorter or often not at all thanks to the new compressed type files.  The tricky thing is that the font-family value basically gets overwritten by other css that is scoped lower, so it ends up that if I set fontFamily: “interBlack” somewhere, that stomps the existing font stack and if jnterBlack is not loaded it will fall back to the browser default.  I’ll add a commit to this to illustrate something I think will work.

PR #101, opened 2023-04-12
Fix invalid tabIndex value

metzsg:
Don't know how that got in there!  React complains like so:
```
react_devtools_backend.js:2655 Warning: Failed prop type: Invalid prop `tabIndex` of type `string` supplied to `ForwardRef(ButtonBase)`, expected `number`.
```


KennethMetz (approved):
Good catch!!

PR #102, opened 2023-04-12
Improve avatar shelf scrolling on slow connections

metzsg:
This change:
- Adds a key to the AvatarIcon, which react lies.  I used the avatar name as the key so that new components will be instantiated on scroll.  Otherwise, the old image will still show while the new image is loading in, making scrolling more confusing.
- Adds a background color to the avatar image itself, which acts as a placeholder while the avatar image is being loaded in.
- Removes an unused prop and key value from inside AvatarIcon.


KennethMetz:
Wow, that delay with the old avatar really was clunky, nice improvement there!!

PR #104, opened 2023-04-13
Fix: Keyboard drag-and-drop functionality

KennethMetz:
- Fixes error thrown when selecting top8 item using 'spacebar' to rearrange list. Keyboard drag-and-drop functionality is restored with this commit.
- Uses new library for drag-and-drop functionality that allows us to KEEP strict mode on....+1 for the dev team!! 🎉🎉🎉🎉🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@emotion/react": "^11.10.5",
         "@emotion/styled": "^11.10.5",
+        "@hello-pangea/dnd": "^16.2.0",
         "@hookform/resolvers": "^2.9.10",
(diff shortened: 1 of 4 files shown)


metzsg (approved):
Nice! I just tried using the spacebar to drag-and-drop and it worked like a charm.  This new package looks like the new fork of the Atlassian one (which isn't maintained anymore), so it is probably good to switch to this one.

- Can we drop the react-beautiful-dnd dep now?
- Should we rebase the PR to add dnd to watchlists after this one iss merged, then switch that one to import from this package also?

KennethMetz:
- yes and
- yes!

PR #106, opened 2023-04-14
Drag handles

KennethMetz:
- Adds drag handles to Top8 and ProfileLists.

- Makes delete button invisible except onHover or when Focused - Note: to get this to work and keep the delete button accessible via the keyboard I could not use the typical `display:none` approach. Instead I make the button invisible by either changing its zIndex or its color...it's kinda hacky, but seems to work.

### DIFF
--- a/src/Components/ProfileListItem.js
+++ b/src/Components/ProfileListItem.js
@@ -1,12 +1,13 @@
 import Box from "@mui/material/Box";
 import ListItemAvatar from "@mui/material/ListItemAvatar";
 import IconButton from "@mui/material/IconButton";
+import Icon from "@mui/material/Icon";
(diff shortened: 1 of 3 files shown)


metzsg (approved):
Nice! I like the addition of the handles.  Regarding visibility and making something keyboard-selectable, I have not checked but would the visibility css property work?

metzsg:
> Nice! I like the addition of the handles. Regarding visibility and making something keyboard-selectable, I have not checked but would the visibility css property work?

Nevermind, just tried with the `visibility` css property but it didn't work.  I think the clever zIndex hack is probably one of our best bets right now to keep this keyboard-accessible.   Very creative!

PR #107, opened 2023-04-14
Adds a description section for each watchlist

KennethMetz:


src/Components/ClickAndEdit.js
```diff
@@ -1,46 +1,35 @@
...
+import { useState } from "react";
 import Box from "@mui/material/Box";
 import TextField from "@mui/material/TextField";
-import Typography from "@mui/material/Typography";
 import Button from "@mui/material/Button";
-import ProfilePageContext from "./ProfilePageContext";
 
-export default function UserBio() {
-  const { profile, isOwnProfile, updateBio } = useContext(ProfilePageContext);
+export default function ClickAndEdit({ data, canEdit, update, index, ml }) {
```
metzsg, line 6: I like the refactor here to make this a general, shareable component.  Nice!

src/Components/ProfileListPage.js
```diff
@@ -149,7 +156,16 @@ export default function ProfileListPage() {
           </Tooltip>
         )}
       </Box>
-
+      {listId.toLowerCase() !== "likes" &&
```
metzsg, line 159: Maybe we could add a variable above called `canEditBio`, similar to `canDelete`?  Then we wouldn't have to duplicate the logic of what it means for a list to be "likes" or "dislikes".

metzsg (approved):
Looks great, thanks for adding this!  Left a couple notes, check them out and see what you think.  I'm cool to merge this as-is, or with those modifications.

src/Components/ProfileListPage.js
```diff
@@ -149,7 +156,16 @@ export default function ProfileListPage() {
...
+        listId.toLowerCase() !== "dislikes" && (
+          <ClickAndEdit
+            data={desc}
+            canEdit={canEdit}
+            update={updateListDesc}
+            index={index}
```
metzsg, line 165: If you want to not need the index dependency here, you could write ClickAndEdit to just take an `onSave` function that has a single string arg with the new value the user is saving.  Then the logic for providing the index in the call to `updateListDesc` could be extracted out here, and the ClickAndEdit component wouldn't have to have any knowledge of indices, and wouldn't really care if an edited value was part of an array or not, since any logic like that would happen outside it.

metzsg: Ex:

Above, 
```js
const onDescSave = (newDesc, index) => updateListDesc(newDesc, index);
```

Then, when creating a ClickAndEdit:
```js
<ClickAndEdit data={desc} canEdit={canEdit} onSave={(newDesc) => onDescSave(newDesc, index)}/>
```

KennethMetz: Epic hack!! 10/10.

PR #108, opened 2023-04-15
Adds Option to Create Private Watchlists

KennethMetz:
Note for when setting up the API to filter out private lists: 
- I'm using the following key to denote whether the list is private or not `privateList: boolean`


metzsg (approved):
Simple and effective!!  I like the icons for public/private.  I'll update the API to filter the private lists ASAP. 

I am now also super motivated to visualize the public/private state on the WatchlistTile for the owner, and to allow the user to toggle this on the WatchlistPage somehow.

PR #110, opened 2023-04-15
Update alpha imports to mui/system/colorManipulator

metzsg:
Looks like there are quite a few alpha-y functions in mui, but the one we had been using for setting alpha values on color hexes is from `@mui/system/colorManipulator`.  Updating to this one fixes some stuff like the cool hover effect on the cover on DetailedView.


KennethMetz (approved):
Good catch!! This was the one mui component I had the hardest time finding when I was getting rid of the top level imports...

PR #113, opened 2023-04-17
Add custom placeholders for ClickAndEdit

metzsg:
This change:
- Allows for placeholders to be specified for `ClickAndEdit.`. This fixes an issue where bios say "Tell us a bit about this list..." when empty.
- Fixes an issue where `onSave` wasn't specified for bios.
- Provides a lighter text color for the placeholder text to better disambiguate it from an actual bio/description/etc.
- Doesn't show any placeholder text if the text is not editable (so the user doesn't see "Tell us about yourself..." on another user's bio page, etc.)

### DIFF
--- a/src/Components/ClickAndEdit.js
+++ b/src/Components/ClickAndEdit.js
@@ -2,8 +2,19 @@ import { useState } from "react";
 import Box from "@mui/material/Box";
 import TextField from "@mui/material/TextField";
 import Button from "@mui/material/Button";
+import useTheme from "@mui/material/styles/useTheme";
 
-export default function ClickAndEdit({ data, canEdit, onSave, ml }) {
+export default function ClickAndEdit({
+  data,
+  placeholder,
+  canEdit,
+  onSave,
+  ml,
+}) {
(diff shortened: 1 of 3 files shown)


src/Components/ClickAndEdit.js
```diff
@@ -22,14 +33,15 @@ export default function ClickAndEdit({ data, canEdit, onSave, ml }) {
...
+            color: data?.length > 0 ? "unset" : theme.palette.text.secondary,
             pb: 1,
             pl: 0,
             ml: ml ? 6 : 0,
             cursor: canEdit ? "pointer" : "unset",
           }}
           onClick={canEdit ? handleDescToggle : undefined}
         >
-          {data?.length > 0 ? data : "Tell us a bit about this list..."}
+          {data?.length > 0 ? data : canEdit ? placeholder : ""}
```
KennethMetz, line 44: Woah...I didn't know you could chain ternaries like this, very cool!!!

KennethMetz (approved):

PR #114, opened 2023-04-17
Add a subtitle to ProfileListPage

metzsg:
It currently has the type of list ("Watchlist", "Watch History" for likes/dislikes) and the number of items in the list. I figure it could be extended to have the number of likes or saves, if we let people like or save watchlists, and other stuff like the top genres in the lists or something. Lmk if you like this idea, we don't have to go with it if it feels redundant or clutter-y.  I do feel like it helps to explain what the page is a bit more, especially if someone lands on the page directly.

### DIFF
--- a/src/Components/ProfileListPage.js
+++ b/src/Components/ProfileListPage.js
@@ -132,6 +132,11 @@ export default function ProfileListPage() {
     marginBottom: "12px",
   };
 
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
Awesome commit!! It really helps the profileListPage look better.

PR #116, opened 2023-04-17
Watchlist - Copy link

KennethMetz:
Changes `X` delete button for watchlists to a drop menu.

Note: now that we will be able to share lists, I thought we should alter our routing to allow guests who are unregistered with firebase to see the content being shared.  Right now, I've only given unregistered guests to see profiles and profile lists.  Lmk what you think we should/shouldn't require folks to register in order to see!


src/Components/ProfilePageContextProvider.js
```diff
@@ -17,7 +17,7 @@ export default function ProfilePageContextProvider({ userId, children }) {
     profile: userOwnsProfile ? localUser : profile,
     profileUserId: userId,
     isOwnProfile: userOwnsProfile,
-    isLoading: loading || profileLoading || !localUser?.uid,
```
metzsg, line 20: I see how this is required to make profiles and lists visible for users without auth, but it does have the side-effect of making the profile page show in a state that looks like there is no data during cold load, between the time the Firebase user is obtained and the LocalUser is populated.  It is usually super short, so not a huge deal, but not as clean.  I have an idea to fix, though.

In general, a shift where we can't assume there is a localUser on all pages that aren't login or landing is definitely a new thing for our site.  I think a lot of decisions on the 'authenticated' pages may have been made in a world where LocalUser is not empty (for example, what happens when the user likes an anime in a Watchlist they've arrived at via a shared link and they don't have a uid?).  I really love the idea that new people can click a shared link and see content immediately without onboarding, and I think it is absolutely essential to grow the site, so I like this change.  But it might be worth thinking about how we want to handle these un-authenticated users.  I like how they get kicked to the login page if they try to explore around, but another option is that they get authed as a guest _automatically_, and can start using the site right away (skipping onboarding, basically).  Then when they click like or create a watchlist, it actually works!  And they still get the upsells to create a permanent account, which can covert them into full users.  This is good as-is for now, but I think this might be worth thinking about.

src/Components/DropMenu.js
```diff
@@ -72,7 +72,7 @@ export default function DropMenu() {
 
   function sendToProfile(e) {
     handleClose(e);
-    navigate(`/profile/${user.uid}`);
+    user ? navigate(`/profile/${user?.uid}`) : navigate("/login");
```
metzsg, line 75: nit: if we know user is non-null due to the ternary, we don't need the null-check later.  This can be just: 
```
user ? navigate(`/profile/${user.uid}`) : navigate("/login")
```

src/Components/Header.js
```diff
@@ -64,7 +64,7 @@ function Header() {
                 <HeaderTab
                   text="Profile"
                   icon={<User />}
-                  path={`/profile/${user.uid}`}
+                  path={user ? `/profile/${user?.uid}` : "/login"}
```
metzsg, line 67: Same as above, can just be:
```
path={user ? `/profile/${user.uid}` : "/login"}
```

src/Components/ProfileListDropMenu.js
```diff
@@ -0,0 +1,141 @@
...
+                  aria-labelledby="composition-button"
+                  onKeyDown={handleListKeyDown}
+                  sx={{
+                    minWidth: "200px",
+                  }}
+                >
+                  <ListItemButton
+                    divider
+                    onClick={() => {
+                      navigator.clipboard.writeText(window.location.href);
```
metzsg, line 107: Nice!  As a follow-up, it might be good to show a Snackbar or something to let the user the copying was successful.

src/Components/ProfileListPage.js
```diff
@@ -161,19 +162,7 @@ export default function ProfileListPage() {
...
-              variant="contained"
-              color="inherit"
-              sx={{ marginLeft: "0.75rem" }}
-              onClick={onDelete}
-            >
-              <X />
-            </IconButton>
-          </Tooltip>
-        )}
+        {canEdit && canDelete && <ProfileListDropMenu onDelete={onDelete} />}
```
metzsg, line 165: In a follow-up, it might be cool to provide this DropMenu for users who aren't the owner as well, with limited features (maybe just Copy Link, later we could add reporting features if we see there is watchlist abuse somehow, etc...).

Before doing this, though, I might want to refactor the variables I made above, for clarity.  For example, it is kinda confusing that `canEdit` indicates that the user has permission to edit this playlist, while `canDelete` doesn't actually indicate that the user has permission to delete it, but rather that it is the type of list that can be deleted.  This is a side-effect of `canDelete` being created before you could view other users' lists, but now I think this can be cleaned up.

metzsg (approved):
Nice! This looks great.  I was just thinking about doing something like this.  I've left a few comments, but there are only a couple of code nits, and the rest are just musings about good follow-ups.  Merge away after you've looked through them 👍

src/Components/ProfilePageContextProvider.js
KennethMetz: Thanks for catching this...I meant to confirm with you that this change didn't have any weird side-effects. Whoops! 😬

And yeah, I agree - that sounds like the best way to go about it.  My only hesitation is we'll blow through loads of guest accounts...but I'm pretty sure Firebase has a way to delete unused/old users, so we can address that issue at a later date.

src/Components/ProfileListDropMenu.js
KennethMetz: Yes!  I was thinking the same thing!  Will incorporate into an upcoming commit.

src/Components/ProfileListPage.js
KennethMetz: Good call - will revise in upcoming commit.

KennethMetz:
Comments have been incorporated!

Note, a dependency has been added: `npm install notistack`

metzsg:
Nice!  I recommend that one, it is great for snackbar stuff

src/Components/RoutingHelper.js
```diff
@@ -53,8 +53,8 @@ export const RoutingHelper = () => {
     );
   } else if (noHeaderMatch) {
     return <Outlet />;
-  } else if (!user && headerMatch) {
-    return <Login />;
+  } else if (!loading && !user && headerMatch) {
+    logInAnon();
```
metzsg, line 57: Nice!  Love how simple this is.  Might also be a good idea to return a BreathingLogo here so that the user sees something while the login happens.  

I know we are pretty confident we won't re-render a bunch here, since this is a super-top-level component, but just to make sure we don't hammer logInAnon on re-renders, it might be nice to wrap the `logInAnon` in a useEffect higher up (can't put a hook inside an if/else), and here you could just return a BreathingLogo.

Maybe:
```js
useEffect(() => {
  if (!loading && !user && headerMatch) {
    logInAnon();
  }
}, [loading, user, headerMatch]);
```

And then:
```js
if (loading && !user) {
  return <BreathingLogo type={"fullPage"} />;
} else if (user && headerMatch) {
  return (
    <>
      <Header />
      <Outlet />
    </>
  );
} else if (noHeaderMatch) {
  return <Outlet />;
} else if (!user && headerMatch) {
  return <BreathingLogo type={"fullPage"} />;
} else {
  return <Outlet />;
}
```

metzsg (commented):
Thanks for these changes!  Things are looking really good.

I am noticing that cold load of a shared link does correctly log-in anonymously, but then PopulateFromFirestore throws because it reads the `user/` document for this new UID and finds nothing there, so it returns undefined as the data, but the rest of that function assumes data is a good object.  Maybe we could add something after `let data = querySnapshot.data();` that checks if data is undefined, and if so, sets it to a default empty LocalUser object for this user's uid.

KennethMetz:
> Thanks for these changes! Things are looking really good.
> 
> I am noticing that cold load of a shared link does correctly log-in anonymously, but then PopulateFromFirestore throws because it reads the `user/` document for this new UID and finds nothing there, so it returns undefined as the data, but the rest of that function assumes data is a good object. Maybe we could add something after `let data = querySnapshot.data();` that checks if data is undefined, and if so, sets it to a default empty LocalUser object for this user's uid.

After looking into this a little, I think the `user/` document is getting returned, but only with (3) keys (uid, authProvider, and name).  So, I updated `PopulateFromFirestore` to add on the remaining keys if they're found to be missing.

PR #117, opened 2023-04-18
Adds like/dislike/add-to-list button to every list item.

KennethMetz:
- Hides these extra buttons except when hovered/focused.

- Added a new background theme to allow the like buttons to access the color, and get them to "disappear".

- Responsive styling will be in a follow-up commit.

### DIFF
--- a/src/Components/DislikeButton.js
+++ b/src/Components/DislikeButton.js
@@ -3,7 +3,7 @@ import Tooltip from "@mui/material/Tooltip";
 import { ThumbsDown } from "phosphor-react";
 import useLikeState from "../Hooks/useLikeState";
 
-export default function DislikeButton({ anime, variant }) {
+export default function DislikeButton({ anime, variant, selected }) {
   const { disliked, setDisliked } = useLikeState(anime);
 
   const onClick = (e) => {
@@ -16,7 +16,7 @@ export default function DislikeButton({ anime, variant }) {
   const size = variant === "contained" ? "large" : "medium";
 
   //Theme colors are taken from the default "palette" theme and MUST have a child color titled "main"
-  const buttonColor = disliked ? "primary" : "inherit";
+  let buttonColor = !selected ? "background" : disliked ? "primary" : "inherit";
 
   return (
     <Tooltip title="Not for me">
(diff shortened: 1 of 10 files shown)


src/Components/LikeButton.js
```diff
@@ -16,7 +16,7 @@ export default function LikeButton({ anime, variant }) {
   const size = variant === "contained" ? "large" : "medium";
 
   //Theme colors are taken from the default "palette" theme and MUST have a child color titled "main"
-  const buttonColor = liked ? "primary" : "inherit";
+  let buttonColor = !selected ? "background" : liked ? "primary" : "inherit";
```
metzsg, line 19: Same as above, can stay `const`

src/Components/DislikeButton.js
```diff
@@ -16,7 +16,7 @@ export default function DislikeButton({ anime, variant }) {
   const size = variant === "contained" ? "large" : "medium";
 
   //Theme colors are taken from the default "palette" theme and MUST have a child color titled "main"
-  const buttonColor = disliked ? "primary" : "inherit";
+  let buttonColor = !selected ? "background" : disliked ? "primary" : "inherit";
```
metzsg, line 19: nit: this variable still won't change in its lifetime (when props change, re-render will run this function again, creating a new variable), so it can remain `const`.

metzsg (approved):
This is looking good!  It is a much better UX to have these buttons there, I love it!!  

I noticed that clicking the Like and Dislike buttons has the desired effect but also takes me to the show page, we probably have to `stopPropagation` in both those buttons' onClick handlers.

We already talked offline about mobile interface, I'm excited to see how that will shape up -- can do in a follow-up PR.

metzsg:
Btw, I share your concern about the user count going way up if we create an anonymous user for every person who visits the site.  Firebase does have an "automatic clean-up" of anonymous users that are older than 30 days: https://firebase.google.com/docs/auth/web/anonymous-auth#auto-cleanup.  Not sure if that is exactly what we need, we might prefer something more like a clean-up of users that have no activity in the last 30 days.  But either way, some work will be required to wipeout user data for these users.  This kind of thing should be done from a server environment, so I will investigate writing a cron that could do this.  (Firestore collections are meant to scale really well, so I am not too worried about having large numbers of localUser documents in the meantime)

PR #118, opened 2023-04-19
Fixes routing behavior

KennethMetz:
-Sends user to /login when clicking EdwardML logo from 'noHeaderRoutes' so it doesn't create a new guest account on accident. -Fixes bug where upon logging out, a guest account would get created as soon as the user was logged out of Firebase due to the sequencing of log out events temporarily satisfying RouterHelper's useEffect (!loading && !user && headerMatch).

### DIFF
--- a/src/Components/DropMenu.js
+++ b/src/Components/DropMenu.js
@@ -86,6 +86,7 @@ export default function DropMenu() {
   }
 
   async function sendToLogout(e) {
+    navigate("/logout");
     await logout();
     setLocalUser({
       likes: [],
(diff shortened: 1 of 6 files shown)


metzsg (approved):
Nice! Both great catches, and these solutions look clean and effective.

metzsg:
Wdyt about having the 'EdwardML' link on `/logout` and `/reset` just take us to `/`, the landing page?  This is what it does on `/` and `/onboarding` so maybe it would be good to be consistent between all signed-out pages.

KennethMetz:
Good thought - that does make a bit more sense to push people toward the landing page.  I'll make these tweaks in an upcoming commit.

PR #119, opened 2023-04-20
Fix colors and clicking on like buttons

metzsg:
At head right now, like buttons on AnimeCards don't have a `selected` value passed in, so their icons are all background-colored.  Also, I noticed clicking on the like/dislike buttons on AnimeCards also navigates to the detail page, so I added back the `e.preventDefault()` to their click handlers.  The 'Add' button already has this, so it was still working and didn't need it.

### DIFF
--- a/src/Components/AnimeCard.js
+++ b/src/Components/AnimeCard.js
@@ -84,7 +84,7 @@ export default function AnimeCard({ anime, large, onChangeSelected }) {
               {anime.display_name}
             </Typography>
             <Box sx={{ display: "flex", justifyContent: "center" }}>
-              <LikeButtons anime={anime} />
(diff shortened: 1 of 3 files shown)


KennethMetz:
Oooff - sorry I missed these. 🤦

I misunderstood the need for both e.preventDefault() and e.stopPropagation().....thought I was double dipping when I was in fact not - lesson learned!

metzsg:
I was actually not clear on it either! I just found [this explanation](educative.io/answers/difference-between-preventdefault-stoppropagation-return-false) that seems pretty understandable.  Apparently, you can also have a click handler return false which will have the same effect as calling both e.preventDefault() and e.stopPropagation().

PR #120, opened 2023-04-20
Responsive list item buttons

KennethMetz:
Creates a drop menu for all the listItem buttons for small screens.

Note, I'm not a fan of the three dots icon I'm using currently...lmk if there's something better out there!


metzsg (approved):
Great change! Love that the mobile and desktop experiences are both really started to be ideal for their devices 🥹

Regarding the icon, maybe try `<DotsThreeVertical />`.  I'll add a commit to swap it in and then merge this branch.  If you don't like it, we can try some other ones

PR #121, opened 2023-04-20
Delete unused components and style sheets

KennethMetz:
Also fixes the like buttons in `/sandbox` by sending it `selected={true}`


metzsg (approved):
Thanks for the cleanup!!  Looks awesome 🤩

PR #122, opened 2023-04-21
Move 'back' button up from beside title to beside avatar

metzsg:
On ProfileListPage, I don't really like how the back button shifted in the title and content below it.  Even worse, we had to add extra logic to `ClickAndEdit`, a general component, to support this extra left margin.  Overall, I like a design where the big titles are always flush with the left side of the content section they describe.  To do this, I:

- Removed the `ml` prop from `ClickAndEdit`
- Moved the back button up to be beside the `ProfileUserBannerSmall` on profile list pages.

As a part of this, I also created a new `BackButton` component.  The new `BackButton` component has an `sx` prop that is piped through to the mui `IconButton` inside, so it can be used to add style props just like with a mui component.  Sadly, it doesn't get the cool autocomplete features when you are filling out the style in vscode, though 😢.

### DIFF
--- a/src/Components/BackButton.js
+++ b/src/Components/BackButton.js
@@ -7,7 +7,7 @@ export default function BackButton({ sx }) {
 
   const onClick = (e) => {
     navigate(-1);
-    e.stopPropogation();
+    e.stopPropagation();
   };
 
   return (
--- a/src/Components/BackButton.js
+++ b/src/Components/BackButton.js
@@ -1,0 +1,18 @@
+import IconButton from "@mui/material/IconButton";
+import { CaretLeft } from "phosphor-react";
+import { useNavigate } from "react-router-dom";
+
+export default function BackButton({ sx }) {
+  const navigate = useNavigate();
+
+  const onClick = (e) => {
+    navigate(-1);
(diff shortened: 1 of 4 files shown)


src/Components/BackButton.js
```diff
@@ -0,0 +1,18 @@
+import IconButton from "@mui/material/IconButton";
+import { CaretLeft } from "phosphor-react";
+import { useNavigate } from "react-router-dom";
+
+export default function BackButton({ sx }) {
+  const navigate = useNavigate();
+
+  const onClick = (e) => {
+    navigate(-1);
+    e.stopPropogation();
```
KennethMetz, line 10: After misspelling this myself countless times, I now know it's e.stopProp**a**gation(), lol.

I'll make this change before merging the branch.

KennethMetz (approved):

KennethMetz:
Great PR!!!  I really like the new location of the back button, and that this change got rid of my hacky `ml` parameter!

src/Components/BackButton.js
metzsg: Oof, my bad! Thanks for the fix

PR #123, opened 2023-04-21
Don't sort likes/dislikes now that the user can arrange them

metzsg:
Since the user can re-arrange the order of their likes and dislikes lists, we shouldn't auto-sort them, or else we risk erasing the user's manual sorting work every time they load a page that calls PopulateFromFirestore 😬.


KennethMetz (approved):
Great catch - thanks for taking care of this!!

PR #124, opened 2023-04-21
Cleanup api calls

KennethMetz:
- Extracts api URL out into a variable that gets defined ONCE for the entire site.  It lives in `APICalls.js`
- Refactors `Search.js` to accommodate above change and adds in loading ghost icons. 
_Note: I'm not in love with using ghost icons during the loading of search results...I almost think using the breathing logo would look better, as we're not guaranteed to get (10) search results back everytime....so lmk what you think!_

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -5,17 +5,13 @@ const fiveMinutesMs = 1000 * 60 * 5;
 const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
 
 export async function APISearch(inputValue) {
-  try {
-    let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
-      mode: "cors",
-    });
-    let responseJson = await response.json();
-    let temp = [];
-    responseJson.items.map((item, index) => temp.push(item));
-    return temp;
-  } catch (error) {
-    console.log(error);
-  }
+  let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
+    mode: "cors",
+  });
+  let responseJson = await response.json();
+  let temp = [];
+  responseJson.items.map((item, index) => temp.push(item));
+  return temp;
 }
 
 export async function APIGetAnime(animeId) {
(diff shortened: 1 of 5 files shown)


src/Components/APICalls.js
```diff
@@ -2,34 +2,27 @@ import { useQuery } from "@tanstack/react-query";
...
-    let temp = [];
-    responseJson.items.map((item, index) => temp.push(item));
-    return temp;
-  } catch (error) {
-    console.log(error);
-  }
+  let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
+    mode: "cors",
+  });
+  let responseJson = await response.json();
```
metzsg, line 11: Should we call `await handleErrors(response);` here, like in other API functions, to make sure that non-OK responses end up throwing?

src/Components/APICalls.js
```diff
@@ -117,23 +107,34 @@ export function useRecommendations(viewHistory) {
...
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(viewHistory),
+      });
       await handleErrors(response);
       let responseJson = await response.json();
       let temp = [];
       responseJson.items.map((item, index) => temp.push(item.anime));
+      console.log(temp);
```
metzsg, line 122: nit: Do we want to commit this log statement?

src/Components/APICalls.js
```diff
@@ -117,23 +107,34 @@ export function useRecommendations(viewHistory) {
...
       return temp;
     },
     { staleTime: fiveMinutesMs, keepPreviousData: true }
   );
 }
+
+export async function getRandomAnimeListing(randomPage, randomItem) {
```
metzsg, line 129: Nice!  This looks great, and I love getting some of the logic of getting the random shelf out from `Home`.

src/Components/SearchGhost.js
```diff
@@ -0,0 +1,30 @@
+import Box from "@mui/material/Box";
+import Container from "@mui/material/Container";
+import Skeleton from "@mui/material/Skeleton";
+
+export default function SearchGhost() {
+  return (
+    <Container maxWidth="sm" style={{ paddingLeft: "16px" }}>
+      <GhostListItem />
```
metzsg, line 8: I like these ghosts and I think they look pretty good.  I think we could maybe try with having fewer -- generally the loading 'ghosts' page doesn't have to line up 1:1 with the rendered page, and this way I think the screen would be less overwhelming, but the user's thought process wouldn't be as interrupted during load repaints as it would be with a loading spinner.  Apparently, UX research says that users perceive ghost pages load faster than spinners (b/c there is more to look at and start parsing while the loading happens, I think) and their user journey feels less interrupted during loading.  Maybe just use 5 here?

metzsg (approved):
Solid PR!  Thanks for the refactor of APICall, that will be helpful.

I left a couple comments, a couple nits and one about the search ghost list items (which I like).  Check em out and merge when ready.

src/Components/APICalls.js
KennethMetz: Great call! Totally forgot we had broken out that error handling function.

src/Components/SearchGhost.js
KennethMetz: Woah - that's really interesting. Sounds good to me!

PR #125, opened 2023-04-21
Fixes typo, making bio render properly

KennethMetz:


src/Components/ProfileSidebar.js
```diff
@@ -48,7 +48,7 @@ export default function ProfileSidebar({ hideDetails }) {
           </Typography>
           <ClickAndEdit
             data={profile?.bio}
-            canEdit={isOwnProfile}
+            isListOwner={isOwnProfile}
```
metzsg, line 51: I'd actually vote to rename the `ClickAndEdit` prop to `canEdit`.  Since it is a general component, and can hypothetically be used to edit text in a list, a profile, a review, etc, the wording should be more oriented toward the desired behavior, and not specific to one application of it.

PR #126, opened 2023-04-24
Prevent unneeded render

KennethMetz:
Prevents a full page render that occurs when the header's Search button is clicked.


metzsg (approved):
Nice fix! This was bothering me but I couldn't easily figure out what it was 😤

PR #127, opened 2023-04-28
Extract anime objects (PRELIMINARY)

KennethMetz:
Preliminary PR that extracts the API anime objects out of localUser and instead fetches them to eliminate size of storage needed in FireStore.


metzsg:
I definitely want to keep looking at this tomorrow, but I have a solution to the issue we were working with today.  The problem is in the following code:
```js
  const ids = profile
    ? [
        ...profile?.likes,
        ...profile?.dislikes,
        ...allAnimeInLists,
        ...profile?.top8,
      ]
    : [];

  const {
    data: animeLists,
    isLoading: isLoading,
    error: error,
  } = useAnimeList(ids);
```

Basically, when there is no profile, we end up with an empty list of ids being passed to `useAnimeList`, and it gets and caches an empty list data response.  Then, when the profile is loaded, while `useAnimeList` is loading the real list of ids, it is still passing the cached empty list out as `data`.  Uncool.

Further, we can't just pretend all empty lists of data from `useAnimeList` are really loading states, because a user with a new profile and no items in any list might load that legitimately. The solution I can think of is to make the 'I have no profile loaded' state obviously different than the 'I have loaded a profile, but I haven't added any items to any lists yet' state.  The easiest way to do that is to make the default `ids` undefined.  And, in `useAnimeList`, if you get an undefined list of ids, return undefined.  Ex:
```js
  const ids = profile
    ? [
        ...profile?.likes,
        ...profile?.dislikes,
        ...allAnimeInLists,
        ...profile?.top8,
      ]
    : undefined;

  const {
    data: animeLists,
    isLoading: isLoading,
    error: error,
  } = useAnimeList(ids);
```
And in `useAnimeList`:
```js
export default function useAnimeList(ids) {
  // To-do: Sort ids

  return useQuery(
    ["anime", ...(ids ?? [])],
    async () => {
      if (ids === undefined) return undefined;
      const response = await APIGetAnimeList(ids);
      return response.items;
    },
    { staleTime: fiveMinutesMs, keepPreviousData: true }
  );
}
```

KennethMetz:
Awesome thinking!!

Implementing this gave me a few errors which I resolved by....
- Making the query key no longer spread `ids`, which threw when undefined.
- Returning `null` instead of `undefined` from useQuery, which was also throwing.

Also, I had to make `ids` remain undefined until profile has a uid...otherwise the blank localUser from localUserContext was causing `useAnimeList` to return an empty array instead of the undefined we want. See below:

````
const ids = profile?.uid
    ? [
        ...profile?.likes,
        ...profile?.dislikes,
        ...allAnimeInLists,
        ...profile?.top8,
      ]
    : undefined;

KennethMetz:
Note, I forgot to comment that on `ProfileListPage` I changed it to only render the subtitle text when we get a valid `items` object.  This causes a little more commotion during rendering, but prevents always having a loading state that will show `0 items` temporarily.

metzsg (approved):
This is looking great!!  I just pushed a commit to this branch that addresses some nits and cleans up some of the unused stuff.  I think we are ready to go.  Let's coordinate on the merge and firestore wipeout, since this will be quite a 💥breaking change💥!  

I am really excited about this one since it sets us up well to scale to handle bigger list sizes, if needed, and add things later like list item pagination or a single anime object cache to prevent fetching redundant list items from `/anime/get`.

PR #129, opened 2023-05-03
Fix error thrown when saving reviews.

KennethMetz:
Also causes reviews to be sorted by date upon saving a new/edited review.


metzsg (approved):
LGTM!

PR #130, opened 2023-05-03
Comments on ProfileListPage

KennethMetz:
This feature allows users to comment on watchlists!!!

A few decisions/assumptions I made that we can tweak if desired:

1.  I've labeled this section **_comments_** despite this section being identical to our **_reviews_** on `DetailedView` 
2. I'm tracking comments in a separate `localUser.comments` field.
3. Currently you can only comment once on a watchlist...this is a bit different than how most comment sections work.
4. Lastly, I currently allow users to comment on their own lists.

### DIFF
--- a/src/Components/DropMenu.js
+++ b/src/Components/DropMenu.js
@@ -96,6 +96,7 @@ export default function DropMenu() {
       bio: null,
       top8: [],
       reviews: [],
+      comments: [],
     });
     handleClose(e);
   }
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -36,6 +36,7 @@ export async function PopulateFromFirestore(user, localUser, setLocalUser) {
 
     if (!data.top8) data.top8 = [];
     if (!data.reviews) data.reviews = [];
+    if (!data.comments) data.comments = [];
     setLocalUser(data);
   } catch (error) {
     console.error("Error loading data from Firebase Database", error);
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -57,6 +57,8 @@ export async function SaveToFirestore(user, localUser) {
           bio: localUser?.bio ?? null,
           top8: [...localUser.top8],
           reviews: [...localUser.reviews],
+          comments:
+            localUser.comments.length > 0 ? [...localUser.comments] : [],
         },
         { merge: true }
       );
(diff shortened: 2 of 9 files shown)


src/Components/Firestore.js
```diff
@@ -57,6 +57,8 @@ export async function SaveToFirestore(user, localUser) {
           bio: localUser?.bio ?? null,
           top8: [...localUser.top8],
           reviews: [...localUser.reviews],
+          comments:
```
metzsg, line 61: To not break existing users who have localUsers without a comments field before they save, we should probably also make sure that the 'comments' field exists in localUser in PopulateFromFirestore.  I'll add a commit to do that.

src/Components/DropMenu.js
```diff
@@ -96,6 +96,7 @@ export default function DropMenu() {
       bio: null,
       top8: [],
       reviews: [],
+      comments: [],
```
metzsg, line 99: Nice! Good to remember this spot.

src/Components/ReviewForm.js
```diff
@@ -65,13 +66,18 @@ export default function ReviewForm({
   });
 
   function populateForm() {
-    if (localUser.reviews.includes(anime.id)) {
-      for (let i = 0; i < animeReviews.length; i++) {
-        if (animeReviews[i].uid === user.uid) {
+    console.log(type);
```
metzsg, line 69: Do we need to commit these test prints?

src/Components/ReviewContainer.js
```diff
@@ -0,0 +1,179 @@
...
+
+import { CaretDown, Minus, Plus } from "phosphor-react";
+import { GetPaginatedReviewsFromFirestore } from "./Firestore";
+import { useEffect, useState } from "react";
+import { useLocation } from "react-router-dom";
+import ReviewFilterDropMenu from "./ReviewFilterDropMenu";
+import ReviewForm from "./ReviewForm";
+import Review from "./Review";
+
+export default function ReviewContainer({ user, docId, type }) {
```
metzsg, line 16: 🔥🔥🔥! I really like extracting this from DetailedView and making it re-usable.

metzsg (approved):
LGTM! Left a couple notes, check them out when you have a sec, then merge!

src/Components/ReviewForm.js
KennethMetz: Haha, no.  No, we do not. 😂

PR #131, opened 2023-05-06
Add more watch buttons for new sites like Hulu, Disney+

metzsg:
In the spirit of not letting perfect be the enemy of good, I've added watch buttons for:
 - Hulu
 - Disney+
 - HBO Max
 - Amazon Prime

And, additionally:
- A 'Google It' button if we have a google kgmid (and also if we have no watch links at all, I will add a "Google It" button with a google search link for the show's title).

It isn't perfect, some shows have two netflix links (different title ids) or two "Google It" buttons (one is for the show entity, another is for the entity that is a specific season of the show).  This could be better in that we could start establishing a priority for urls, so we'd link people to the better one only, but that improvement can come later.


src/Components/UrlButtons.js
```diff
@@ -22,23 +28,41 @@ export default function UrlButtons({ anime }) {
...
-            <UrlButton title="Crunchyroll" link={url} image={crunchyroll} />
-          )}
-          {url.includes("funimation") && (
-            <UrlButton title="Funimation" link={url} image={funimation} />
-          )}
-          {url.includes("netflix") && (
-            <UrlButton title="Netflix" link={url} image={netflix} />
-          )}
-        </Fragment>
+      {buttonData.map((data) => (
```
KennethMetz, line 31: Tassssssty!!!!!! 🔥🔥🔥🔥🔥

src/Components/UrlButtons.js
```diff
@@ -22,23 +28,41 @@ export default function UrlButtons({ anime }) {
...
     </Box>
   );
 }
 
+// Maps a list of url strings to a list of button data objects:
+// {url, title, image}.  Returns undefined for unsupported urls.
+function getButtonData(urls) {
+  return urls
+    .map((url) => {
+      if (url.includes("crunchyroll.com")) {
```
KennethMetz, line 43: I don't mind the duplicates for the time being...any savvy user should understand the difference easily enough ¯\_(ツ)_/¯

src/Components/UrlButtons.js
```diff
@@ -1,18 +1,24 @@
...
+
+  // Add a simple google link if there are none.
+  if (!buttonData.length) {
+    const encodedName = encodeURIComponent(anime.display_name);
+    const googleUrl = `https://www.google.com/search?q=${encodedName}`;
+    buttonData.push({ url: googleUrl, title: "Google It", image: google });
+  }
 
   return (
     <Box
```
KennethMetz, line 24: Not sure if you're still cooking on this shelf, but currently it gunks up our mobile view if there's 4+ urlButtons :(

Just adding a `flexWrap: "wrap"` here does a good enough job (if you're not already implementing buttons!)

KennethMetz (approved):
Awesome stuff man!!! It's definitely a nice improvement!

src/Components/UrlButtons.js
metzsg: Oh, snap!  Nice catch -- I didn't even check that 😧.  Fix applied.

PR #132, opened 2023-05-07
Let search bar expand to full width on mobile

metzsg:
Right now, our header search bar is kind of small on mobile, which isn't terrible, but then the autocomplete panel beneath it is also very small, and this seems like a lost opportunity to make that thing the width of the whole screen, like search in mobile apps usually is.  See rotten tomatoes' mobile web search for an example of the bar expanding to full width on focus.  I don't do any cool animations here, nor do I provide a 'cancel' button to dismiss the search bar, but those could come as future improvements.


KennethMetz:
Looks great man - it's a real nice improvement!

PR #134, opened 2023-05-08
Fix/bugs in comments

KennethMetz:
Cleans up some mess in the 'comments' section.
- Corrects wording of tooltips for 'comments' vs 'reviews'
- Allows comments to be deleted
- Keeps 'see more' button from rendering on page load.  

Note: I need to clean-up the 'see more' button behavior further.  Currently, if the # of reviews is divisible by 4 it won't hide the 'see more' button until it's clicked once more.  Will fix in an upcoming PR.

### DIFF
--- a/src/Components/Review.js
+++ b/src/Components/Review.js
@@ -255,7 +255,7 @@ export default function Review({
             marginTop: "10px",
           }}
         >
-          <Tooltip title={`Applaude this ${type}`} followCursor>
+          <Tooltip title={`Applaude this ${typeSingular}`} followCursor>
             <div>
               <EmojiReactionChip
                 docId={docId}
@@ -270,7 +270,7 @@ export default function Review({
             </div>
           </Tooltip>
 
-          <Tooltip title={`Love this ${type}`} followCursor>
+          <Tooltip title={`Love this ${typeSingular}`} followCursor>
             <div>
               <EmojiReactionChip
                 docId={docId}
@@ -285,7 +285,7 @@ export default function Review({
             </div>
           </Tooltip>
 
-          <Tooltip title={`Disagree with this ${type}`} followCursor>
+          <Tooltip title={`Disagree with this ${typeSingular}`} followCursor>
             <div>
               <EmojiReactionChip
                 docId={docId}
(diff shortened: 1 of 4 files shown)


src/Components/Review.js
```diff
@@ -253,7 +255,7 @@ export default function Review({
             marginTop: "10px",
           }}
         >
-          <Tooltip title="Applaude this review" followCursor>
+          <Tooltip title={`Applaude this ${type}`} followCursor>
```
metzsg, line 258: Hmm, if `type` is not singular, would this end up being 'Applaud this comments'?  Same below for love and disagree buttons.

src/Components/ReviewContainer.js
```diff
@@ -79,9 +79,7 @@ export default function ReviewContainer({ user, docId, type }) {
             </Typography>
           )}
           {!showReviewForm ? (
-            <Tooltip
-              title={type === "Comments" ? "Add a comment" : "Add a review"}
-            >
+            <Tooltip title={`Add a ${type}`}>
```
metzsg, line 82: Would this end up being "Add a comments" or "Add a reviews" with a plural noun?

src/Components/ReviewForm.js
```diff
@@ -43,13 +43,15 @@ export default function ReviewForm({
 
   let edited = false;
 
+  const typeSingular = type === "comments" ? "comment" : "Review";
```
metzsg, line 46: nit: Is it intentional to capitalize 'Review' but not 'comment'?  If not, which casing do you prefer?

src/Components/ReviewForm.js
```diff
@@ -201,7 +203,7 @@ export default function ReviewForm({
           }}
         />{" "}
         <TextField
-          label="Review"
+          label={typeSingular[0].toUpperCase() + typeSingular.slice(1)}
```
metzsg, line 206: Clever!

metzsg (approved):
Thanks for the fixes!   Left a couple notes about the noun plurality a few places.  I probably wouldn't have even thought to check it but when I saw the `typeSingular` that got me thinking about it!

src/Components/Review.js
KennethMetz: Lolololol.  Great catch!

PR #138, opened 2023-05-15
WIP: Testing out new search experiences

metzsg:
This is a WIP branch -- not for merging!  I am just creating this PR so we can have a place to discuss these new ideas. Checkout this branch and try out the new search experiences and play around with them when you have some time.  

I am curious if you have any thoughts about them, and whether they feel like interesting avenues to keep exploring.  I think there is a cool new feature in here, especially with the semantic-search backend looking through reviews, but I don't think I have it quite dialed in yet.  Looking for good ideas for ways we can create a new 'discovery' type experience to find new shows based on some criteria, especially if that criteria is "I want a relaxing show to watch after a stressful day" or "I want find out what are must-watch classics that I may have missed"

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -2,7 +2,8 @@ import { useQuery } from "@tanstack/react-query";
 
 const fiveMinutesMs = 1000 * 60 * 5;
 
-const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
+//const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
+const apiUrl = "https://api-jet-dev-lfoguxrv7q-uw.a.run.app";
 
 export async function APISearch(inputValue) {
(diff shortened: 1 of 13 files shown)


KennethMetz:
Woah, very cool!!! This feature has got LEGSSSS! It's a lot of fun to play around with. Also this paired with the "Data from Edward" helps figure out if a show's worth checking out super fast! 

A couple of my thoughts are below:
- Without direction, this could be confusing if users are expecting an OpenAi sort of interaction. It seems like some kind of guidance/guardrails could help users understand it's searching user reviews and make the most of that.

A couple of layout ideas:

- Have it be a header link, maybe "ask Edward" or "ask Anime Fans".  This could expand into a search bar...but also have a Question Mark icon that can take users to a Dialog/pop-up that gives more direction on the differences?

- Alternatively, the header link could just be generic..."Advanced Search" which goes to its own page that's dominated by a search bar, but also explains "good" prompts to give some handholding.

I'll keep thinking on this though!

PR #140, opened 2023-05-26
Editable Display Names

KennethMetz:
Allows display names to be changed.  Also the following:
- Re-styles <ClickAndEdit> to look less shitty
- Note, long text entries can break our layout...will add truncation to our to-do list.

### DIFF
--- a/src/Components/ClickAndEdit.js
+++ b/src/Components/ClickAndEdit.js
@@ -22,12 +22,19 @@ export default function ClickAndEdit({
   const [editDesc, setEditDesc] = useState(false);
   const [editedDesc, setEditedDesc] = useState(undefined);
 
+  // TO-DO:
+  // Add in ghost cards for ProfileUserBanner, and only render ClickAndEdit
+  // when data has been loaded in.  Then below useEffect can
+  // be deleted and useState(data) for editDesc.
+
   useEffect(() => {
     setEditedDesc(data);
   }, [data]);
   const smallScreen = useMediaQuery(theme.breakpoints.down("sm"));
 
-  //To-do: Provide truncated name/description for lengthy entries so they don't break layout.
+  // TO-DO:
+  // Provide truncated name/description for lengthy entries so they
+  // don't break layout.
 
   function saveDesc() {
     onSave(editedDesc);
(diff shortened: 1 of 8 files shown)


src/Components/ClickAndEdit.js
```diff
@@ -16,7 +20,14 @@ export default function ClickAndEdit({
   const theme = useTheme();
 
   const [editDesc, setEditDesc] = useState(false);
-  const [editedDesc, setEditedDesc] = useState(data);
+  const [editedDesc, setEditedDesc] = useState(undefined);
+
+  useEffect(() => {
```
KennethMetz, line 25: I have no idea why useState(data) worked for populating editedDesc for the "bio" section, but doing the same thing for "display Name" always gave me undefined....😠

Anyways, this was my workaround.

src/Components/ClickAndEdit.js
```diff
@@ -27,76 +38,95 @@ export default function ClickAndEdit({
...
-            }}
-          ></TextField>
-          <Button
-            size="small"
-            variant="outlined"
-            color="inherit"
-            onClick={(e) => {
-              saveDesc();
-              handleDescToggle();
+  if (data) {
```
metzsg, line 41: I can change my name to an empty string, which then won't allow me to edit further since the empty string is falsy.  Should we be more specific and check `if (data !== undefined)` perhaps?

src/Components/ClickAndEdit.js
metzsg: Nice, this definitely seems like it will work.  With `useState` the initial value is trapped on the first evaluation, and must are first loading this component before the user profile is loaded in, so `data={profile?.name}` evaluates to undefined. 

Bio worked fine because we don't render the main profile page while the profile data is still loading.  See:
```
  if (isLoading) {
    return <ProfileMainPageGhost />;
  }
```
in `ProfileMainPage`.

By contrast, `ProfileUserBanner` doesn't have any such alternate ghost-display that is shown while the data deps are still loading.  If we add that, we wouldn't need to do this type of move bc `ClickAndEdit` would only be created once the data deps are present.  It might be a good thing to do anyways, since the 'loading' sstate in ProfileUserBanner doesn't quite match the ghost cards for the rest of the profile page.  Maybe leave a todo here to remove this trick after `ProfileUserBanner` has a ghost display implemented.

metzsg (approved):
Looks good, man!  A solid feature.

Left a couple comments, check them out when you have a sec, then when you are ready, merge away.  I will rebase my typography PR on top of the changes from this PR once this one is merged.

src/Components/ClickAndEdit.js
metzsg: Also, we should be able to get rid of this check once we apply the fix I mentioned to implement ghost display for `ProfileUserBanner`.  Maybe leave a TODO?

KennethMetz:
Great catches!!! I appreciate the review/input!

PR #141, opened 2023-05-29
Set up fallback fonts, move all typography to theme

metzsg:
Ok this one was a bit of a pain, but this change sets up all font data to be specified through the theme generated by theme.js. This is much more maintainable, as fonts are no longer hardcoded in individual components. Also, fallback fonts are always applied (previously, if we set `fontFamily: interMedium` on a component, that became the whole font stack for that component, and the fallbacks go back to being system default, probably Times. Yuck.).  So now the "Flash Of Unstyled Text" on page load before the custom fonts load looks much nicer (basically a tweaked Arial is used instead of Inter).

This change touches a _lot_ of files, but 90% of this change is either:
- Deleting lots of `fontFamily: interMedium` since that is now default everywhere, including in material components
- Changing headers from `h3` to `<Typography variant="h3">`
- Replacing some `fontFamily: interExtraBold` type things with `fontWeight: 800`, where we just wanted bold text

This change _should_ be almost entirely a visual no-op. Lmk if you see major changes in appearance and I can fix them.

I moved to loading all font variants under the same font name "Inter" and using font weights to select the bold variations. So, if you need to make some text use SemiBold, just use `fontWeight: 600`, etc. No more need to hardcode font names. Here are the weights mapped to font variants:
- interMedium: default, 500
- interSemiBold: 600
- interExtraBold: 800
- interBlack: 900

When creating headers, use the appropriate semantic header, when possible. There is a meaningful structure to headers `h1` through `h5`, which can be seen in /sandbox, but I will also list them here:

- h1: Hero titles in Montserrat, as seen on the landing page.  Used very rarely.
- h2: Page titles. Heavy weight (Black/900) and large, usually only one of these at the top of every page.
- h3: Section headers. Appear over a subsection of a page. Also heavy (Black/900), not as large. Seen extensively on DetailedView, etc.
- h4: Light headers. For shelf headers under a section header, as seen on Home. Font is _almost_ the same size as h3 but uses a SemiBold weight (600)
- h5: Subheader. Basically body text, but using a heavy (Black/900) weight.

Using only these will also help ensure consistency in fonts, weights, and sizes between pages in the app.

These can be accessed by using `<Typography variant="h3" />` or similar. It bums me out that material doesn't override the base header tags `<h3>`, but oh well. It was a pain to chase down all the headers and convert them to Typography, but we only have to do it once.

A cool thing about this implementation, in addition to better centralizing all font stuff and making sure the default font stack is applied everywhere, is it makes it easy to experiment with large typography changes for the app (want to see what happens if we use a different font for body text? what happens if section headers are less bold? This should be a one or two line change).

### DIFF
--- a/src/Components/AddToListDropMenu.js
+++ b/src/Components/AddToListDropMenu.js
@@ -199,7 +199,7 @@ export default function AddToListDropMenu({ anime, variant, selected }) {
                       primary="Add to Watchlists"
                       primaryTypographyProps={{
                         fontSize: "1rem",
-                        fontFamily: "interExtraBold",
+                        fontWeight: 800,
                       }}
                     />
                   </ListItemButton>
@@ -209,7 +209,7 @@ export default function AddToListDropMenu({ anime, variant, selected }) {
                       tabIndex={-1}
                       sx={{
                         margin: "10px 0px 10px",
-                        fontFamily: "interSemiBold",
+                        fontWeight: 600,
                       }}
                       secondaryAction={
                         <AddButtonForTop8 anime={anime} list={localUser.top8} />
@@ -230,7 +230,6 @@ export default function AddToListDropMenu({ anime, variant, selected }) {
                             tabIndex={-1}
                             sx={{
                               margin: "10px 0px 10px",
-                              fontFamily: "interMedium",
                             }}
                             secondaryAction={
                               <AddButton anime={anime} list={item.name} />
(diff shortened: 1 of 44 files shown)


KennethMetz:
Yeeesh - what a headache!! Lesson definitely learned about using themes for fonts EARLY ON.  

Thanks for knocking it out. This will also make code splitting later on even easier.

KennethMetz:
Also, thanks for updating the sandbox  - that's super helpful!

KennethMetz (approved):

PR #143, opened 2023-05-30
Add ghost ProfileUserBanner

metzsg:
As discussed in our previous PR.  This will prevent the banner from rendering its contents until profile data is fully loaded, so the ClickAndEdit initial value can be trapped correctly.


KennethMetz (approved):
Awesome - thanks for taking care of this one!!

PR #144, opened 2023-05-31
Choose handle

KennethMetz:
This PR adds the ability to select a handle. It will allow "anonymous" users to view the site without one, but forces all registered user to select one prior to being able to peruse the site.  Currently it is not possible to change handles, we can change this quite easily in the future if desired.

A new collection has been made to store the normalized handles. Note, transactions will be implemented in a follow-up PR to prevent the "race-off" problem previously discussed.


src/Components/HandleDialog.js
```diff
@@ -0,0 +1,173 @@
...
+    });
+    navigate("/");
+  }
+
+  function handleFormSubmission() {
+    let isDuplicate;
+    newLocalUser = { ...localUser };
+    newLocalUser.handle = handle;
+    CheckForHandleDuplicates(handle, user)
+      .then((result) => {
```
metzsg, line 62: style note: since functions that return a promise (including 'then' functions) provide their return values to the next function as the argument, we could use that mechanism to provide `isDuplicate`, there is no need for the external local variable.  something like:
```js
CheckForHandleDuplicates(handle, user)
.then((isDuplicate) => {
  if (isDuplicate) {
    // oh no!
  } else {
    // yay!
  }
});
```

metzsg:
I am really liking this feature!! I tried it out and it worked nicely.  Thanks for working on it.  I left a note or two about the handle-claiming system, because I think that could be tightened up a bit

src/Components/ProfileUserBanner.js
```diff
@@ -85,6 +86,7 @@ export default function ProfileUserBanner() {
               fontSize: { xs: "1.66rem", md: "2.5rem" },
             }}
           />
+          <Typography variant="h4">@{profile?.handle}</Typography>
```
metzsg, line 89: do we want an "@" sign to show up for users without a handle? or for this to just be blank?

src/Components/Firestore.js
```diff
@@ -201,3 +202,38 @@ export async function SaveListReactionsToFirestore(docId, updatedRxns) {
...
+
+export async function CheckForHandleDuplicates(
+  handle,
+  user,
+  newLocalUser,
+  setLocalUser
+) {
+  try {
+    const collectionRef = collection(db, "handles");
+    const q = query(collectionRef, where("handle", "==", handle));
```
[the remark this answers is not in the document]
metzsg: Ok, just realized I was wrong about the first part -- the user *shouldn't* be able to query the `/users` collection for other users with the handle since they shouldn't be able to read `/users` documents for anyone other than themselves, since these contain private data.  So the current way you have it does work around that issue.  But if we do have intention of preventing those races at some point, we probably should still key `/handles` by normalized handle.  You can do that with or without the transaction in this PR, whichever you prefer.

src/Components/HandleDialog.js
KennethMetz: Oooweeee that's clean!!! Thanks for the correction.

src/Components/Firestore.js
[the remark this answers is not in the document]
KennethMetz: Awesome feedback - I like the idea of "future-proofing" this kind of stuff right now, so I included the transactions.

src/Components/ProfileUserBanner.js
KennethMetz: Great catch!!! @ symbol has been removed if there's no handle to display

metzsg (approved):
Looking good.  I added a commit to fix up the error handling and display, and have tested that out and it is working nicely.  I'll make a follow-up PR to fix the name-stomping bug on registration once this is merged in.

PR #145, opened 2023-05-31
Fixes bugs/nits

KennethMetz:
- Shows correct display name in `DropMenu.js`
- Fix display name in `ProfileUserBanner` to spread across entire page when opened for editing.
- Fix bug where if a `ClickAndEdit` component is open on your profile for editing, then after navigating to another user's profile would cause the wrong display name to show.

### DIFF
--- a/src/Components/ClickAndEdit.js
+++ b/src/Components/ClickAndEdit.js
@@ -6,6 +6,7 @@ import useTheme from "@mui/material/styles/useTheme";
 import Tooltip from "@mui/material/Tooltip";
 import Box from "@mui/material/Box";
 import useMediaQuery from "@mui/material/useMediaQuery";
+import { useLocation } from "react-router-dom";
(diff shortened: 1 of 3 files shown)


src/Components/ClickAndEdit.js
```diff
@@ -17,8 +18,9 @@ export default function ClickAndEdit({
 }) {
   placeholder = placeholder ?? "Enter some text...";
 
+  const location = useLocation();
   const theme = useTheme();
-
+  console.log(location);
```
metzsg, line 23: remove test log?

metzsg (approved):
Thanks for the fixes!  

If we end up with more profile->profile component lifetime bugs, I wonder if we should set some high-level key on profile pages to keep components from being re-used if the user navigates from one user's profile to another's.  Just something to think about

PR #146, opened 2023-06-03
Implement fallback font overrides for less layout shift

metzsg:
Which I learned about at https://developer.chrome.com/blog/font-fallbacks.  Inter is a bit bigger than Arial, so I adjusted Arial to be larger for less layout shift once Inter loads.  What I was doing before with actual text css properties was wrong.

Before: https://drive.google.com/file/d/1RK10nygHeuNcCt2O4qOfuelKCcH7AssS/view?usp=sharing
After: https://drive.google.com/file/d/1PkwJqAiXQqW_EhFhj0AVkrKuzwEYA_lE/view?usp=sharing

I also added a Roboto fallback since Android doesn't have Arial installed by default.

### DIFF
--- a/src/Components/theme.js
+++ b/src/Components/theme.js
@@ -1,7 +1,7 @@
 import createTheme from "@mui/material/styles/createTheme";
 import { alpha } from "@mui/system/colorManipulator";
 
-const mainFontStack = "Inter, Arial-fallback, sans-serif";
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
Awesome work - thanks for doing the research on this!!  The transition between fonts is much more palatable now.

PR #147, opened 2023-06-04
Lighthouse improvements

KennethMetz:
A couple of improvements based on Lighthouse feedback.
- Gives edward logo a fixed ratio to prevent re-paints
- Allows montserrat font to use a fallback font...(I realize we don't have a fallback font specified for montserrat at the moment - we can discuss more at our next mtg!)


metzsg (approved):
Looks good! I like the change to force 'swap' mode for the montserrat fonts

PR #149, opened 2023-06-07
Search caching & debouncing

KennethMetz:
- Changes our API requests to search into a TanStack query for caching purposes.
- Debounces search requests, currently pauses for 450 ms prior to firing request to API...that feels about right to me, but feel free to tinker with the value.
- Prevents misleading "No Options" result from being shown in popper during loading state.

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -15,6 +15,9 @@ export function useAPISearch(inputValue) {
       let responseJson = await response.json();
       let temp = [];
       responseJson.items.map((item, index) => temp.push(item));
+      // Clears option field if search term is "".
+      if (inputValue.length === 0) return [];
+      // Otherwise return query results
       return temp;
     },
     { staleTime: fiveMinutesMs }
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -4,6 +4,23 @@ const fiveMinutesMs = 1000 * 60 * 5;
 
 const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
 
+export function useAPISearch(inputValue) {
+  return useQuery(
+    ["search" + inputValue],
+    async () => {
+      let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
+        mode: "cors",
+      });
+      await handleErrors(response);
+      let responseJson = await response.json();
+      let temp = [];
+      responseJson.items.map((item, index) => temp.push(item));
+      return temp;
+    },
+    { staleTime: fiveMinutesMs }
+  );
+}
+
 export async function APISearch(inputValue) {
   let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
     mode: "cors",
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -6,7 +6,7 @@ const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
 
 export function useAPISearch(inputValue) {
   return useQuery(
-    ["search" + inputValue],
+    ["search:" + inputValue],
     async () => {
       let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
         mode: "cors",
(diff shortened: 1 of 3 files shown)


src/Components/APICalls.js
```diff
@@ -4,15 +4,24 @@ const fiveMinutesMs = 1000 * 60 * 5;
...
+export function useAPISearch(inputValue) {
+  return useQuery(
+    ["search:" + inputValue],
+    async () => {
+      let response = await fetch(`${apiUrl}/anime/search?query=${inputValue}`, {
+        mode: "cors",
+      });
+      await handleErrors(response);
+      let responseJson = await response.json();
+      let temp = [];
```
metzsg, line 16: If the point of this code with `temp` is to return an array of all the items in `responseJson.items`, could this all be replaced with the following code?
```js
if (inputValue.length === 0) {
  return [];
}
return responseJson.items;
```

src/Components/TitleAutocomplete.js
```diff
@@ -35,15 +45,32 @@ export default function TitleAutocomplete({ search, setShowSearch }) {
     if (key === "Escape") setShowSearch(false);
   }
 
+  // Prevents popper from showing "No Results" during loading state and while there's no search term
+  useEffect(() => {
+    if (!searchLoading && inputValue.length > 0) openPopper();
+    else closePopper();
+  }, [searchOptions]);
+
+  const debouncedSearch = useCallback(
```
metzsg, line 54: nice!

src/Components/APICalls.js
```diff
@@ -4,15 +4,24 @@ const fiveMinutesMs = 1000 * 60 * 5;
...
+      responseJson.items.map((item, index) => temp.push(item));
+      // Clears option field if search term is "".
+      if (inputValue.length === 0) return [];
```
metzsg, line 19: Do we want to return nothing on the 0-prefix (empty) search?  The api will return a list of the most popular anime, might be nice to have it show up as suggestions before the user types anything.

src/Components/TitleAutocomplete.js
```diff
@@ -35,15 +45,32 @@ export default function TitleAutocomplete({ search, setShowSearch }) {
     if (key === "Escape") setShowSearch(false);
   }
 
+  // Prevents popper from showing "No Results" during loading state and while there's no search term
+  useEffect(() => {
```
metzsg, line 49: do we want a set of options when the user clicks in the box with no query?  Also, do we want the popper to flash off and on again all the time as they add to their query and more requests get sent?  I might suggest keeping the popper open the whole time.  Instead, maybe use `keepPreviousData` in the useQuery options for search queries (to keep old results up while a new query is in flight), and you can then use `searchLoading` to detect the very first load before any data has been returned.  I haven't actually tried this out, so maybe there are bugs I am not predicting, but I feel like this would be a pretty smooth experience that would match other search boxes (google, yt, etc)

src/Components/TitleAutocomplete.js
```diff
@@ -35,15 +45,32 @@ export default function TitleAutocomplete({ search, setShowSearch }) {
...
+    if (!searchLoading && inputValue.length > 0) openPopper();
+    else closePopper();
+  }, [searchOptions]);
+
+  const debouncedSearch = useCallback(
+    debounce((value) => {
+      const normalizedSearchTerm = value?.toLowerCase();
+      setSearchTerm(normalizedSearchTerm); // Triggers APISearch call by changing TanStack query key
+    }, 450),
```
metzsg, line 58: I like a interval around 200ms, personally -- feels more snappy but still prevents multiple requests if I am typing fast.

metzsg (commented):
Looks good man!  Thanks for doing this one.  Left a couple comments, just some product-type questions around if we should show a zero-prefix set of suggestions, and if we should try to keep the popper open while the user is active in the search box.  I also think I like a shorter debounce interval of ~200ms.  Try it and lmk what you think!

src/Components/APICalls.js
KennethMetz: Squeaky clean - very nice!!

src/Components/APICalls.js
KennethMetz: I thought that was confusing...but it looks like YT/LinkedIn/Discord all agree with you, so I've incorporated it back in.

src/Components/TitleAutocomplete.js
KennethMetz: I implemented essentially what you describe.  Note, I added in a 'loading' state for the popper when useQuery is making its call. 

Also, to eliminate a couple of unwanted "flashes" I (1) use the data from useQuery **_directly_** in our Autocomplete component, and (2) Combined the useQuery 'isFetching' state into one loading variable to rule them all!

metzsg (approved):
This is looking great!!  I love the loading state with the circular loading indicator in the search field 🤌.  Let's merge

metzsg:
FYI, now that this is in, I've pushed an API update so anime searches come from the new index.  This means we will match on the description contents as well as titles -- try searching 'ninja' to see it in action

PR #150, opened 2023-06-07
Show 'For You' during localUser loading state

KennethMetz:
Change /home to show the 'For You' ghost cards while localUser is loading, regardless of whether the user has liked shows required to generate the 'For You' content.

Great catch on this one Steve - this really smooths out the page load feel.  😎


metzsg (approved):
Yes!!! This one is a huge improvement to the feel of /home

PR #151, opened 2023-06-07
Display privacy setting

KennethMetz:
- Shows a globe icon for public lists and a lock icon for private lists on list pages the user owns.
  Note - currently a lists privacy setting is permanent.  This could certainly be changed at a later date if desired.
- Adds an orange reminder snackbar when user copy the link for a private list.
- Changes successful snackbar color to green to provide a more familiar UX.
- Tweaked styling of main profile pages to accommodate long display/handle names.

### DIFF
--- a/src/Components/ProfileListDropMenu.js
+++ b/src/Components/ProfileListDropMenu.js
@@ -125,7 +125,6 @@ export default function ProfileListDropMenu({
                         variant: "success",
                         style: {
                           fontSize: "0.9rem",
-                          // background: theme.palette.primary.main,
                         },
                         anchorOrigin: {
                           vertical: "top",
@@ -182,7 +181,6 @@ export default function ProfileListDropMenu({
                             variant: "success",
                             style: {
                               fontSize: "0.9rem",
-                              // background: theme.palette.primary.main,
                             },
                             anchorOrigin: {
                               vertical: "top",
--- a/src/Components/ProfileListDropMenu.js
+++ b/src/Components/ProfileListDropMenu.js
@@ -22,6 +22,7 @@ export default function ProfileListDropMenu({
   deletableList,
   userId,
   listId,
+  privateList,
 }) {
   const [open, setOpen] = useState(false);
   const anchorRef = useRef(null);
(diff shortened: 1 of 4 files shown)


src/Components/ProfileListPage.js
```diff
@@ -329,3 +340,18 @@ function getItemsText(items) {
 
   return `${items.length} items`;
 }
+
+function GetPrivacySymbol({ privateList }) {
```
metzsg, line 345: super nit: functions usually have names that start with verbs, but components are usually just nouns.  i.e. this could be "PrivacySymbol", then used in JSX would be `<PrivacySymbol>`, similar to other things like `<Button>` or `<Grid>`.  (unless this is the symbol for "GetPrivacy", which is an action or something).

src/Components/ProfileListPage.js
```diff
@@ -329,3 +340,18 @@ function getItemsText(items) {
...
+  return (
+    <Tooltip title={privateList ? "Private list" : "Public list"}>
```
metzsg, line 347: can we just wrap the tooltip around the icon?

src/Components/ProfileListDropMenu.js
```diff
@@ -119,11 +120,12 @@ export default function ProfileListDropMenu({
                     onClick={() => {
                       navigator.clipboard.writeText(window.location.href);
                       setOpen(false);
+
                       enqueueSnackbar("Link copied to clipboard", {
                         variant: "success",
                         style: {
                           fontSize: "0.9rem",
-                          background: theme.palette.primary.main,
+                          // background: theme.palette.primary.main,
```
metzsg, line 128: can we delete this commented-out line?

src/Components/ProfileListDropMenu.js
```diff
@@ -161,7 +182,7 @@ export default function ProfileListDropMenu({
                             variant: "success",
                             style: {
                               fontSize: "0.9rem",
-                              background: theme.palette.primary.main,
+                              // background: theme.palette.primary.main,
```
metzsg, line 185: can we delete this commented-out line?

metzsg (approved):
LGTM! Left a couple style nits.  I love the use of the icon, and how it matches the privacy setting control when creating a list.

src/Components/ProfileListPage.js
KennethMetz: Great catch!! Thanks for keeping me honest.

Also, I'm sure you'll never guess what this component started as 😂😂😂

PR #152, opened 2023-06-10
Create new useAuthActions hook, use for Login and Register

metzsg:
This change fixes the name-stomping bug, and attempts to reduce complexity of the login/registration code flow.  I do this by making a new hook that will hold all app business logic for logging in and registering, with an effort to minimize code re-use and to be easy to reason about.

The hook will login, register, or link using the appropriate `firebase/auth` method, then:

For all successful authentications (login/register/link):
 - Writes uid, authProvider, and email to `/users`

Then, for registrations/links:
 - Copies the firebase User displayName to localUser.name, and
 - Writes existing onboarding LocalUser to `/users` 

Finally, for all successful authentications:
 - Loads LocalUser from `/users`

The login and user data logic will be separate from the UI logic (showing loading indicators, error text, etc), which will reside in the Login and Register components themselves.  This will make both the hook and components more decoupled and easily testable.  (Tests to come in a follow-up).

The `Login` and `Register` components will then have a single submit handler function which simply calls the right authAction function, waits until it is done, then interprets and displays any errors, or else navigates the user to /home.

I have manually tested it with registering, linking, and logging in and things *seem* to be working nicely.  As a follow-up, I'd like to add some automated tests that mock out the firebase auth and firestore methods, so that we can run them and assure ourselves every login/registration method is fully functional, without creating tons of accounts while doing so.

I'd also like to add a loading indicator to the Login page, and a more universal indicator to Register that will appear when registering with Google/Twitter.

### DIFF
--- a/src/Components/Login.js
+++ b/src/Components/Login.js
@@ -1,17 +1,7 @@
-import React, { useEffect, useState, useContext } from "react";
+import React, { useState } from "react";
 import { Link, useNavigate } from "react-router-dom";
-import {
-  auth,
-  logInWithEmailAndPassword,
-  signInWithGoogle,
-  signInWithTwitter,
-} from "./Firebase";
-import { useAuthState } from "react-firebase-hooks/auth";
 
 import "../Styles/Login.css";
-import { PopulateFromFirestore } from "./Firestore";
-
-import { LocalUserContext } from "./LocalUserContext";
 
 import Box from "@mui/material/Box";
 import Container from "@mui/material/Container";
@@ -24,13 +14,14 @@ import useTheme from "@mui/material/styles/useTheme";
 import TwitterIcon from "@mui/icons-material/Twitter";
 import google from "../Styles/images/google.svg";
 import EdwardMLLogo from "./EdwardMLLogo";
+import useAuthActions from "../Hooks/useAuthActions";
 
 export default function Login() {
-  const [localUser, setLocalUser] = useContext(LocalUserContext);
+  const authActions = useAuthActions();
 
   const [email, setEmail] = useState("");
   const [password, setPassword] = useState("");
-  const [user, loading, error] = useAuthState(auth);
+
   const navigate = useNavigate();
   const theme = useTheme();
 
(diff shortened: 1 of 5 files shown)


metzsg:
Note: I moved this logic from Firebase.js to a hook so I could access LocalUser and its setter without having to pipe them around.

src/Components/Register.js
```diff
@@ -89,13 +79,47 @@ export default function Register() {
...
-    }
   }, [user, loading, forwardToken]);
 
+  const handleRegister = async (provider) => {
+    try {
+      setRegisterError(null);
+      if (provider === "google") {
+        !user
+          ? await authActions.registerWithGoogle()
+          : await authActions.linkWithGoogle();
```
KennethMetz, line 90: With this logic it's possible to overwrite an existing account - I don't like this as our /register and /login screens look identical, so it seems like an easy mistake to make that could wipe out a bunch of work.

See my suggestion below!

src/Hooks/useAuthActions.js
```diff
@@ -0,0 +1,149 @@
...
+
+    // If this is registration, write onboarding LocalUser to /users.
+    if (isRegistration) {
+      // Set initial display name first.
+      const newLocalUser = { ...localUser };
+      newLocalUser.name = displayNameOverride ?? user.displayName;
+      setLocalUser(newLocalUser);
+      await SaveToFirestore(user, newLocalUser);
+    }
+
```
KennethMetz, line 52: The above SaveToFirestore will wipe an existing google/twitter account.  However, I see it's also the function that prevents the name stomping issue.

Perhaps, we add a second conditional statement, that only is truthy if the google/twitter user credentials are new to us.  I would like to query Firebase Authentication to confirm this....however, I can't figure it out rn. So my workaround would be querying Firestore instead.  (Note, while we can search for AuthProvider and Email for Google accounts, we would need to add in the TwitterHandle when registering Twitter accounts for this approach to work.

Also, I like the idea of throwing a snackbar to alert users if the account already exists, so they know their onboarding selections got scrapped.

src/Components/Login.js
```diff
@@ -46,16 +37,38 @@ export default function Login() {
...
 
-  useEffect(() => {
-    if (loading) {
-      // trigger a loading screen?
-      return;
-    }
-    if (user) {
-      PopulateFromFirestore(user, localUser, setLocalUser);
+  const login = async (provider) => {
+    try {
```
KennethMetz, line 41: Ooohh, very nice!!

KennethMetz (approved):
Gnarly - looks good to me!!  

I left a few notes about the possibility of overwriting existing Google/Twitter authed accounts with the current logic.  Not sure whether this was an intended outcome or not.

src/Hooks/useAuthActions.js
metzsg: Hmm, very valid concerns.  I'm trying to make sure I understand the scenario where existing accounts could be wiped.  Are you considering a case where I am logged in using email/password, for example, then I go to register page and click 'log in with google' or something?  Because, in this case, it would seem like the LocalUser from the previous account would overwrite the new one, and that does seem bad.  Good catch.

I just added a commit that will silently log the user in to the existing account, in this case.  I check for existence by checking for the /users document, which seems like a pretty safe way to check this.  Take a look and let me know what you think. An alternative would be to fail a registration in this case -- I could do that easily as well, and show the user a message like "This identity already has an account.  Try logging in instead."  Not sure which is better.

metzsg: Actually, as I think about it, in the case that the user chooses an already existing account, it could go two ways:

1) The user was not already authenticated -- here, my new code would take effect and basically log the user in to that account, throwing away their onboarding data.

2) The user was already authenticated -- here, they will attempt to link the existing account with this google/twitter account.  This may not succeed if the existing account is not anonymous.  If it is anonymous, but they are trying to link with an identity that already has a different uid, Idk exactly what happens.  I *think* it should fail.  We should verify this, though...

KennethMetz: Haven't dug into the changes yet, but here's how to recreate the account wiping I noticed.  
1. Make an account using Google/Twitter and populate it. 
2. Sign out (I refresh too)
3. Go through onboarding and register an account using Google/Twitter.
4. All info in account from Step 1 will be wiped.

KennethMetz (approved):
Awesome - that existingUserDoc check works perfect for the account wiping I noticed!  Nice fix!!  

Also, I like the logic you implemented, as this will prevent an account with LOTS of lists, reviews, reactions, etc. from getting wiped while only wiping out a few likes from onboarding.

KennethMetz:
Bug: LocalUser from /onboarding wasn't being saved.

Fix: I pushed through a tweak of our syntax of the check of whether an account already exists. Just needed to unwrap the promise!!

PR #153, opened 2023-06-12
Add fallback font for montserrat

KennethMetz:
- Adds 'sans-serif' as a fallback font for montserrat where it was missing (like in our logo!!)
- Cleans up code (swaps in <EdwardMLLogo> component a couple of places, fixes the header on 404, deletes unused component)


metzsg (approved):
Nice! A solid change.  Love the increase in component re-use over repetition.

PR #154, opened 2023-06-13
User search

KennethMetz:
Incorporates search for user's displayNames and handles into both the autocomplete component and the search page.

Future Task:
- Add 'see more' buttons, to allow for more than 10 titles and 10 users to be displayed on the search page.

### DIFF
--- a/src/Components/Search.js
+++ b/src/Components/Search.js
@@ -50,6 +50,7 @@ export default function Search() {
     setSearch(location.state);
   }, [location]);
 
+  // Locates where to place the "USERS" banner
   useEffect(() => {
     setIndexOfUsers(undefined);
     for (let i = 0; i < searchResults.length; i++) {
--- a/src/Components/Search.js
+++ b/src/Components/Search.js
@@ -1,6 +1,6 @@
 import "../Styles/Search.css";
 import { useNavigate, useLocation } from "react-router-dom";
-import { useState, useEffect, useContext } from "react";
+import { useState, useEffect, useContext, useMemo, Fragment } from "react";
 import Box from "@mui/material/Box";
 import ListItemAvatar from "@mui/material/ListItemAvatar";
 import Divider from "@mui/material/Divider";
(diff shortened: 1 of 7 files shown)


src/Components/Search.js
```diff
@@ -30,19 +32,35 @@ export default function Search() {
...
+  const loadingSearch = searchLoading || searchFetching;
+
+  const [indexOfUsers, setIndexOfUsers] = useState();
 
   useEffect(() => {
     setSearch(location.state);
   }, [location]);
 
+  // Locates where to place the "USERS" banner
+  useEffect(() => {
```
metzsg, line 54: Clever!  

If you are using `useEffect` here just to make a computation only when a variable changes, you could also consider `useMemo`:

```js
const indexOfUsers = useMemo(() => {
  for (let i = 0; i < searchResults.length; i++) {
    if (searchResults[i].uid) {
      return i;
    }
  }
  return undefined;
}, [searchResults]);
```

The function will only be run when searchResults changes.  Otherwise, `useMemo` will have stored the previous result and will just return that.

metzsg (approved):
Looks great!  Works well, and I like the section headers in the autocomplete box to show both titles and users in the same results.  This really makes the site more explorable, I love it

src/Components/Search.js
KennethMetz: Ahh, that's much better!! Great suggestion!

src/Components/TitleAutocompleteOptions.js
```diff
@@ -0,0 +1,46 @@
+import Box from "@mui/material/Box";
+import Avatar from "@mui/material/Avatar";
+import { useMemo } from "react";
+import { getAvatarSrc } from "./Avatars";
+import Typography from "@mui/material/Typography";
+import ListItemText from "@mui/material/ListItemText";
+
+export default function RenderedOptions({ props, options }) {
```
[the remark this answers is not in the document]
KennethMetz: Great catches! Changed to `TitleAutocompleteOption` so it shows up next to its parent in the File Explorer.

PR #156, opened 2023-06-15
Bug Fixes in <HandleDialog.js>

KennethMetz:
- Eliminates warning being thrown for having a `<div>` nested in a `<p>` (actually it was a `<DialogContentText>`, but they're treated as `<p>` components unless explicitly stated).
- Prevents overwriting display names for guest users who had chosen one prior to registering their accounts.
- Fixes styling


metzsg (approved):
Another good one!  Good thinking on being more restrictive about updating the display name to match the new handle.

PR #158, opened 2023-06-20
Fix review bugs

KennethMetz:
- Fixes the 'edited' text on reviews/comments to only show on content that was actually edited.
- Displays new reviews/comments immediately on the page after submittal.
- Fixes the total count of reviews/comments that got bungled when we switched over to paginated reviews.

**NOTE:** The 3rd commit contains the following bug:
- The total # of comments always returns as (0), while the same code used for retrieving the # of reviews returns the correct number.

Also, FYI I left in an unnecessary `getDocs()` request in `GetReviewCount` to show the collection ref being used is correct.

### DIFF
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -13,6 +13,7 @@ import {
   endBefore,
   where,
   runTransaction,
+  getCountFromServer,
 } from "firebase/firestore";
 import { db } from "./Firebase";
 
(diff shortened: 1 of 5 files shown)


metzsg:
That aggregation issue with the comment counts looks like a bug in firestore.  It doesn't respond well to `+` symbols in the collection path, since these end up in the api url that it requests, unescaped.

KennethMetz:
Great catch!!! The `+` symbol has been deleted and the aggregation issue resolved.

KennethMetz:
Note: This breaks all list comments made thus far.  Once you give your blessing and merge, I'll go through and wipe the old data from Firestore.

metzsg (approved):
Looks great!  Just confirming this will also update what is written in the `comments` field of LocalUser?

Re: the wipeout or update to roll out this change, it will be a little tricky since these document ids are denormalized into LocalUser in that comments list.  At this point, might be simplest to just go through users/ and find the localUsers that have data in that field and just manually delete or update that field, based on what approach you choose.

After you merge this and clear or update the existing docs in watchlistData/ and users/, it would also be a good idea to push a new build of the web app, just so there aren't any old versions out there that would still write the old formats

metzsg:
Note: leaving unmerged so you can merge, update the db, and push builds all at the same time.

PR #161, opened 2023-06-30
Restyle popper components

KennethMetz:
This PR makes the below styling changes to our poppers:
1. Truncates very long list/handle/display names to prevent the layout from breaking.
2. Gives poppers a maxWidth so all contents are visible on mobile.
3. Makes each list name 'clickable' in `AddToListDropMenu` to navigate directly to each list.

### DIFF
--- a/src/Components/DropMenu.js
+++ b/src/Components/DropMenu.js
@@ -137,7 +137,7 @@ export default function DropMenu() {
         role={undefined}
         placement="bottom-start"
         transition
-        style={{ zIndex: "4" }}
+        style={{ zIndex: "4", maxWidth: "320px" }}
       >
         {({ TransitionProps, placement }) => (
           <Grow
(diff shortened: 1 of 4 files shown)


src/Components/DropMenu.js
```diff
@@ -179,12 +179,25 @@ export default function DropMenu() {
                         color: user ? "primary" : theme.palette.text.primary,
                       }}
                       primary={localUser?.name ? `${localUser?.name}` : "Guest"}
-                      primaryTypographyProps={{ fontWeight: 600 }}
+                      primaryTypographyProps={{
+                        variant: "h3",
```
metzsg, line 183: Nice! Good use of typography variants

metzsg (approved):
Good updates!  I like the list titles as links, and the hover effect really makes it intuitive.  Also like removing the screaming caps for top 8.  Merging!

PR #162, opened 2023-06-30
Bug Fix: Fallback for anime images

KennethMetz:
Bug: When the url for an anime image is broken (as is the case for Bleach Thousand Year Blood War) its tile does not show up in a few spots.

Fix: Makes all places that show the image for an anime also have a gray background color.  This allows anime with broken image links to still have a visible tile (albeit with a gray background...but we can dress this up later!)


metzsg (approved):
Thanks or the fix! 

And regarding the missing image, we've already picked up the new url for this show's image, I just need to sync it to prod.

PR #164, opened 2023-07-01
Removes itemId from Chip in GenreChipsNew

KennethMetz:
- React flags this as an invalid DOM property and suggests 'itemID' instead. However, I don't think this parameter is needed so I deleted it, but feel free to add it back in if I missed something.


metzsg:
Looks good! I added it due to a comment I saw in an example at https://www.npmjs.com/package/react-horizontal-scrolling-menu, but react doesn't like it and it looks like the actual scrolling shelf code will also use `key` props as the id for items, so we should be all good to remove it

PR #165, opened 2023-07-02
Add react-query for /anime/id/similar, use staleTime for /analyze

metzsg:
Two improvements.

1. `useSimilarAnime` now uses react-query.
2. `useAnimeAnalysis` now has a stale time, which cuts down on extra unneeded requests, especially when navigating away from a DetailedView page.


KennethMetz:
Looks good!  Nice work catching that unnecessary api call!

PR #166, opened 2023-07-04
Notification ghosts

KennethMetz:
Adds notification ghosts and a "global" loading state for each notification item.

This required some re-factoring - all hooks used to load the info for a notification line were moved to be on the same level so a global loading state could be utilized.

### DIFF
--- a/src/Components/NotificationListItem.js
+++ b/src/Components/NotificationListItem.js
@@ -33,7 +33,7 @@ export function NotificationListItem({ item, handleClose, index }) {
   if (item.docType === "reviews") notificationType = "emojiOnReview";
   if (item.docType === "list") notificationType = "emojiOnList";
 
-  let ignoreCall = notificationType === "emojiOnReview" ? false : true;
+  let ignoreCall = notificationType !== "emojiOnReview";
 
   const { data: interactorData } = useProfile(item.interactorId);
   const { data: listOwnerData, loading: loadingListOwner } = useProfile(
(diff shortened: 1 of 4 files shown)


src/Hooks/useAnime.js
```diff
@@ -21,7 +20,7 @@ export default function useAnime(animeId, cachedAnime) {
     // Otherwise, fetch anime from API.
     setFetchedAnime(undefined);
     setError(undefined);
-    APIGetAnime(animeId)
+    APIGetAnime(animeId, ignoreCall)
```
metzsg, line 23: it would be cool if we could do this without calling the APIGetAnime function.  maybe something up above near the other early-out.  consider something like: 
```js
    // Don't load if we have the right data already.
    if (getMatchingAnime(animeId, [cachedAnime, fetchedAnime])) {
      return;
    }
    // Don't send API request if told to ignore it.
    if (ignoreCall) {
      return;
    }
```

src/Components/NotificationListItem.js
```diff
@@ -19,27 +20,61 @@ import useMediaQuery from "@mui/material/useMediaQuery";
...
-  const [showGhosts, setShowGhosts] = useState(false);
+
+  // Determine what type of notification this is, in order to select text output
+  let notificationType;
+  if (!item.docType) notificationType = "listComment";
+  if (item.docType === "comments") notificationType = "emojiOnComment";
+  if (item.docType === "reviews") notificationType = "emojiOnReview";
+  if (item.docType === "list") notificationType = "emojiOnList";
+
+  let ignoreCall = notificationType === "emojiOnReview" ? false : true;
```
metzsg, line 36: nit: probably could be cleaner without the ternary.  the result of a boolean comparison is already a bool.  something like `const ignoreCall = notificationType !== "emojiOnReview"`

metzsg (approved):
Looks good! A very clever refactor.  

Added a quick note about a way to incur a little less tech debt in terms of the `ignoreCall` arg.  I do intend to take the `useAnime` api call and wrap that in react-query, maybe once I do that I can think of some way to avoid needing it all together, but for now it works well and helps us manage the loading state regardless of the type of dependencies we have for the list item.

KennethMetz:
Great comments!!

Also, I figured out how to get rid of that clunky `ignoreCall` term by raising the conditional outside of useAnime!🎈 🎈 🎈

metzsg:
Looks good! I like the new approach with the null animeId better.  Nice work

PR #167, opened 2023-07-05
Add dynamic <head> tags

metzsg:
I've installed a package called `react-helmet` that allows us to dynamically change the html doc head tags for each page. I've leveraged it to:
1. Update the `theme-color` which allows the iOS system bg color to match our bg color.  [before](https://drive.google.com/file/d/16LgmVQPVqj5qZ5ftYnqnanhOAP6U-nyl/view?usp=sharing), [after](https://drive.google.com/file/d/1abNQnRYEXTuMRo3bvDFCrNhrByTCLRPo/view?usp=sharing).
2. Set different page titles, depending on the page.  Now the titles will be things like "Naruto - EdwardML", "Your Profile - EdwardML", "Search Results - EdwardML".  This will make the back stack look nicer. [before](https://drive.google.com/file/d/1G5jJ1IC3CottD4qPvvC-ssSQB6z2WKC9/view?usp=sharing), [after](https://drive.google.com/file/d/1hCorAyZvxa64NbRPMUBRLRkTvoYL5Afv/view?usp=sharing).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-firebase-hooks": "^5.1.1",
+        "react-helmet": "^6.1.0",
         "react-hook-form": "^7.42.1",
         "react-horizontal-scrolling-menu": "^4.0.4",
         "react-router-dom": "^6.4.5",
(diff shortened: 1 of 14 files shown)


KennethMetz (approved):
Great PR!!! Definitely helps it look more finished - and in a super simple PR to boot! 💯

PR #168, opened 2023-07-06
Update useAnime to use react-query

metzsg:
I didn't earlier, since 99% of the time this hook wouldn't do any fetching since all DetailedView pages were linked with anime data already in the route state.  Now that notifications uses this hook as well, it seems more common that this will be called without a valid cachedAnime and will have to make an API call, so I've moved it over to react-query.  This should save us redundant api calls, and offer all the other nice react-query benefits.


KennethMetz (approved):
LGTM!!!

PR #169, opened 2023-07-11
Notifications - Refactored Infinite Scroll

KennethMetz:
Updates notification popper to work as follows:
- Initial notifications read from real time listener.
- Real time listeners are used to trigger the "new notifications badge" and the initial (4) notifications only.
- All additional notifications are read using static `getDocs` - so once popper is opened all displayed notifications are "frozen in time". 
- Any new notifications will only be shown by clicking a "New Notifications" button that will appear the same time as the "new notifications badge".

This update significantly reduces the number of Firestore reads required by the notification popper....especially when a user goes way back in time.

Outstanding Item: After clicking on a notification, the user loses their spot among all of the loaded notifications.

### DIFF
--- a/src/Components/NotificationDropMenu.js
+++ b/src/Components/NotificationDropMenu.js
@@ -38,13 +38,12 @@ export default function NotificationDropMenu() {
 
   const [
     notifications,
-    setNotifications,
     showMore,
     hideBadge,
-    GetNotis,
-    setLastVisible,
+    getNotis,
     showNewNotiButton,
     displayLatestNotis,
+    resetPagination,
   ] = useContext(NotificationsContext);
 
   const isMobileWidth = useMediaQuery(
@@ -52,7 +51,7 @@ export default function NotificationDropMenu() {
   );
 
   const handleToggle = () => {
-    if (!open) GetNotis();
+    if (!open) getNotis();
     setOpen((prevOpen) => !prevOpen);
   };
 
@@ -158,10 +157,7 @@ export default function NotificationDropMenu() {
         {({ TransitionProps }) => (
           <Grow
             {...TransitionProps}
-            onExited={() => {
-              setNotifications([]);
-              setLastVisible();
-            }}
+            onExited={() => resetPagination()}
             style={{
               transformOrigin: "right top",
             }}
@@ -200,7 +196,7 @@ export default function NotificationDropMenu() {
                       <ListItemButton
                         sx={{ display: "flex", justifyContent: "center" }}
                         onClick={() => {
-                          GetNotis();
+                          getNotis();
                           setLoadingMoreNotis(true);
                         }}
                       >
--- a/src/Components/NotificationDropMenu.js
+++ b/src/Components/NotificationDropMenu.js
@@ -1,4 +1,11 @@
-import React, { useContext, useEffect, useMemo, useRef, useState } from "react";
+import React, {
+  Fragment,
+  useContext,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import ClickAwayListener from "@mui/material/ClickAwayListener";
 import Grow from "@mui/material/Grow";
 import Paper from "@mui/material/Paper";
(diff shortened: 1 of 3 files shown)


src/Components/NotificationDropMenu.js
```diff
@@ -158,55 +158,83 @@ export default function NotificationDropMenu() {
         {({ TransitionProps }) => (
           <Grow
             {...TransitionProps}
+            onExited={() => {
+              setNotifications([]);
```
metzsg, line 162: Added a note about this section below in the NotificationsProvider.

src/Components/NotificationsProvider.js
```diff
@@ -91,8 +157,10 @@ export default function NotificationsProvider(props) {
         setNotifications,
         showMore,
         hideBadge,
-        moreNotiRequests,
-        setMoreNotiRequests,
+        GetNotis,
+        setLastVisible,
```
metzsg, line 161: I notice that `setNotifications` and `setLastVisbile` are both exposed by this interface, but using them correctly requires a knowledge of the implementation of this context.  There might be better encapsulation here if those methods stayed internal implementation details.  I notice that they are only used to reset the state after the popper is exited.  Perhaps the NotificationsContext could instead expose a function `resetPagination` or something like that, which calls those methods and resets the pagination in the right way.  Then, other devs (aka me) can use NotficationsContext without having to understand how it works under the hood.

ex, in NotificationsProvider:
```js
function resetPagination() {
  setNotifications([]);
  setLastVisible();
}

// and then NotificationsContext.Provider could just return `resetPagination`
```

and in NotificationDropMenu:
```js
const [
    notifications,
    showMore,
    hideBadge,
    GetNotis,
    showNewNotiButton,
    displayLatestNotis,
    resetPagination,
  ] = useContext(NotificationsContext);


<Grow onExited={() => resetPagination()} />
```

src/Components/NotificationsProvider.js
```diff
@@ -2,67 +2,133 @@ import { useEffect, useState } from "react";
...
+      if (latestNoti.length === 5)
+        setLastVisibleListener(
+          documentSnapshots.docs[documentSnapshots.docs.length - 2]
+        );
+      else setLastVisibleListener(null);
+    });
+    return unsub;
+  }, [user]);
+
+  async function GetNotis() {
```
metzsg, line 53: style-wise, is there a specific reason this function starts with a capital letter?  that is usually reserved for classes, or react components/contexts/etc, whereas standard functions usually get `lowerCamelCase`

metzsg (approved):
Really well done!  I messed around with this one quite a bit and it seems very robust.  I think this is a great UX, and much more modern feeling than viewing one page at a time.  The 'new notifications' popup stands out well, and matches existing design patterns.  Nice work on this one!

Added a couple comments, check em out and lmk if you have any questions

src/Components/NotificationsProvider.js
KennethMetz: Great comment - thanks for outlining the advantages to keeping the methods internal!

src/Components/NotificationsProvider.js
KennethMetz: Nope, just overlooked it - great catch!!

src/Components/NotificationsProvider.js
```diff
@@ -2,67 +2,133 @@ import { useEffect, useState } from "react";
...
-      limit(5 + moreNotiRequests * 5)
-    );
-    const unsub = onSnapshot(q, (querySnapshot) => {
-      const notis = [];
-      querySnapshot.forEach((doc) => {
-        notis.push({ ...doc.data(), firestoreDocId: doc.id });
+    const q = query(notisRef, orderBy("time", "desc"), limit(5));
+    const unsub = onSnapshot(q, (documentSnapshots) => {
+      const latestNoti = [];
+      documentSnapshots.forEach((doc) => {
```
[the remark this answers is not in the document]
KennethMetz: Ooooh, very slick!!! 💯 💯 💯

PR #170, opened 2023-07-11
Expands avatar shelves on big screens

KennethMetz:


metzsg (approved):
Nice!  I like the way this looks/works in the app much better.  That ternary is huge, but really quite readable, I like it.

PR #171, opened 2023-07-14
Community Lists on Home

KennethMetz:
Displays a randomly chosen user list on the home page.  

The methodology used for the random firestore document selection came from this [stack overflow article](https://stackoverflow.com/questions/46798981/firestore-how-to-get-random-documents-in-a-collection) we discussed earlier.

Notes: 
1. I put a limit on the number of times the function that picks the random list `getRandomCommunityList()` will re-run if it picks an empty list. I did this to guard against an infinite loop of firestore reads that would cook our quota 😂, however, this means that it's possible for NO list to be chosen. (Thinking aloud here - alternatively, we could hard code in a fallback list?)
1. I also limited the # of anime pulled from the random list (to 24) in order to cap the # of reads this shelf could cost.
1. Currently, only the corresponding `watchlistData` document gets deleted when users delete a list.  Both sub-collections `reactions` and `reviews` are left in-place as deleting collections from a web client is [discouraged by Google](https://firebase.google.com/docs/firestore/manage-data/delete-data#collections).  I like the idea of running a script every so often to delete all sub-collections that no longer have their ancestor document, just to keep things tidy. ✨

### DIFF
--- a/src/Components/AnimeShelfClassic.js
+++ b/src/Components/AnimeShelfClassic.js
@@ -7,7 +7,7 @@ import useMediaQuery from "@mui/material/useMediaQuery";
 import useTheme from "@mui/material/styles/useTheme";
 
 import { CaretLeft, CaretRight } from "phosphor-react";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useSwipeable } from "react-swipeable";
 import AnimeCard from "./AnimeCard";
 
@@ -67,6 +67,11 @@ export default function AnimeShelfClassic({ items }) {
     onSwipedRight: (eventData) =>
       setStartIndex(Math.max(startIndex - itemsPerPage, 0)),
   });
+
+  // Resets pagination back to beggining whenever new anime is received
+  useEffect(() => {
+    setStartIndex(0);
+  }, [items]);
 
   return (
     <Box
--- a/src/Components/CommunityListShelf.js
+++ b/src/Components/CommunityListShelf.js
@@ -33,7 +33,6 @@ export default function CommunityListShelf({ data, anime, titleStyle }) {
   return (
     <Fragment>
       {/* Render list name */}
-
       <Box sx={{ display: "flex", alignItems: "center" }}>
         <Typography
           variant="h4"
@@ -66,8 +65,7 @@ export default function CommunityListShelf({ data, anime, titleStyle }) {
       <Box
         tabIndex="0"
         onKeyDown={(e) => {
-          if (e.key === "Enter")
-            navigate(`/profile/${profile.uid}/list/${data.listId}`);
+          if (e.key === "Enter") navigate(`/profile/${profile.uid}`);
         }}
         sx={{
           display: "inline-flex",
--- a/src/Components/CommunityListShelf.js
+++ b/src/Components/CommunityListShelf.js
@@ -1,0 +1,99 @@
+import Typography from "@mui/material/Typography";
+import Box from "@mui/material/Box";
+import { getAvatarSrc } from "./Avatars";
+import { Fragment, useEffect, useMemo, useState } from "react";
+import { useNavigate } from "react-router-dom";
+import { useProfile } from "./APICalls";
+import Avatar from "@mui/material/Avatar";
+
+import AnimeShelf from "./AnimeShelf";
+import Skeleton from "@mui/material/Skeleton";
+
+export default function CommunityListShelf({ data, anime, titleStyle }) {
+  const navigate = useNavigate();
+  const { data: profile, isLoading: profileLoading } = useProfile(data?.userId);
+  const [listName, setListName] = useState();
+
+  const avatarSrc = useMemo(
+    () => getAvatarSrc(profile?.avatar),
+    [profile?.avatar]
+  );
+
+  useEffect(() => {
+    if (!profile?.lists) return;
+    for (let item of profile.lists) {
+      if (item.id === data.listId) setListName(item.name);
+    }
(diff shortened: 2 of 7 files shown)


src/Components/Firestore.js
```diff
@@ -303,3 +319,69 @@ export async function MarkNotificationsSeenOrRead(notiArray, IdToNotify, verb) {
     }
   }
 }
+
+export async function getRandomCommunityList() {
+  // Repeat call to try and ensure a list with at least (1) anime title is found.
+  for (let i = 1; i < 6; i++) {
+    try {
+      const randomNumber = Math.floor(Math.random() * 9999999);
```
metzsg, line 327: maybe make the random number generator into a private function in this file -- the implementation has to be exactly the same here and above for this method to work, so it might make sense to just have one implementation

also, I think 10 million (technically, 9,999,999) is a safe number -- hopefully we should have an algorithmic watchlist recommendation system well before we get anywhere close to saturating that keyspace.

src/Components/Firestore.js
```diff
@@ -303,3 +319,69 @@ export async function MarkNotificationsSeenOrRead(notiArray, IdToNotify, verb) {
...
+
+  return runTransaction(db, async (transaction) => {
+    // This code may get re-run multiple times if there are conflicts.
+    const listDoc = await transaction.get(listDocRef);
+
+    // The deletion will only succeed if the document read above has not changed since then.
+    // Otherwise, it will re-run this function.
+    transaction.delete(listDocRef);
+  });
+  // To-do: Write script to delete all sub-collections WITHOUT an ancestor.
```
metzsg, line 384: I can do this.  Once that exists, we may want a 'deleted' flag that we can set on the watchlistData so it is easy to query just those that are marked for deletion, which will reduce unnecessary reads.

src/Components/AnimeShelfClassic.js
```diff
@@ -68,6 +68,11 @@ export default function AnimeShelfClassic({ items }) {
       setStartIndex(Math.max(startIndex - itemsPerPage, 0)),
   });
 
+  // Resets pagination back to beggining whenever new anime is received
+  useEffect(() => {
```
metzsg, line 72: Nice! Thanks for the fix.  This had been bugging me.

src/Components/CommunityListShelf.js
```diff
@@ -0,0 +1,103 @@
...
+  const coarsePointer = useMediaQuery("(pointer: coarse)");
+
+  const isTouchscreen = cannotHover && coarsePointer;
+
+  const avatarSrc = useMemo(
+    () => getAvatarSrc(profile?.avatar),
+    [profile?.avatar]
+  );
+
+  const listName = useMemo(() => {
```
metzsg, line 26: Nice use of `useMemo`!

metzsg (approved):
Looking really good!  I like the display for this on Home, it really brings a lot of life and dynamic-ness to the page.  The ghost version looks cool too!  Implementation is solid and easy to read.  Good separation of concerns.  Not a lot of comments on this one!

src/Components/Firestore.js
KennethMetz: Great thinking!!

I now only delete the `random` field so it won't get picked to show up on the home page, and I add `awaitingDeletion: true` to the doc.  Let me know if you'd like this flagging mechanism to look/act differently!

src/Components/CommunityListShelf.js
KennethMetz: Thanks! Writing this function really helped me understand it better. Here `useMemo` is definitely much easier to read/understand than having to chase around a separate state!

src/Components/Firestore.js
KennethMetz: Could you elaborate on the benefit of using a private function here?  I'm not understanding how this would generate the random # any differently.

metzsg: it wouldn't generate the number any differently, but it would guarantee the number generation happening during read and write are done the same way.  I imagine if you used different numbers in each location, it would break the system.  e.g. if you used 9999999 during write, but 99999999 during read (can you spot the difference?), you'd get the same watchlist like 90% of the time.  

instead, either consider a const for the number, like:

```js

const watchlistRandomMax = 9999999;

// then later...

// on document creation

const random = Math.floor(Math.random() * watchlistRandomMax);

const data = {
  ownerId: "whoever",
  listId: "whatever",
  random: random
};
setDoc(db, watchlistRef, data);

// and on read, as well:

const random = Math.floor(Math.random() * watchlistRandomMax);

```

or else the whole generation in a function, like:

```js

function generateRandomWatchlistInt() {
  return Math.floor(Math.random() * 9999999);
}

// then, later...

// on document creation

const data = {
  ownerId: "whoever",
  listId: "whatever",
  random: generateRandomWatchlistInt()
};

await setDoc(db, watchlistRef, data);

// and on document reads...

const random = generateRandomWatchlistInt()

```

this makes the consistency between these implementations explicit, instead of implicit.

but an optional thing, up to you if you want to apply this. making the number const is quick and easy and probably good practice, just to make sure it is the same.

src/Components/Firestore.js
metzsg: Great improvements, that will work nicely.  One wipeout syncer coming right up...

src/Components/Firestore.js
KennethMetz: Okay, I see. Yes, that's a very good point! Thanks for the explanation!!!

Will incorporate prior to merging.

PR #172, opened 2023-07-16
Add a 'Trending' shelf to Home

metzsg:
I've added in the first 'explore more' position, but we can move it elsewhere if you'd like.

I am working on ranking this (and other) shelves so hopefully will have some more changes soon that continue to add more dynamic and personalized content to Home.

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -98,6 +98,12 @@ export function useAnimeMH() {
   );
 }
 
+export function useAnimeTN(selectedGenre) {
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
Great stuff man! This is a fantastic shelf - really makes the whole site feel more "up to date"...also being able to filter this shelf by genre is next level!!!

Also, I like this shelf being first in the 'explore more' line up.  And speaking of order, we can move the fan lists shelf wherever.  I don't have a strong opinion on where it should land.

PR #174, opened 2023-07-19
Disables Emojis for Guests

KennethMetz:
- When guests react to content, the notification it creates won't display properly due to guests not having a handle.  
- I think this is also a good step to prevent feature abuse (folks creating oodles of guest accounts to pad stats).


metzsg (approved):
LGTM!

PR #175, opened 2023-07-19
Hides guest watchlists from the Home shelf.

KennethMetz:
- Prevents guest lists from being displayed on home. 
- Once a guest registers their account, all their lists then get marked as eligible to be displayed on home.

### DIFF
--- a/src/Hooks/useAuthActions.js
+++ b/src/Hooks/useAuthActions.js
@@ -13,6 +13,7 @@ import {
 import { useContext } from "react";
 import { LocalUserContext } from "../Components/LocalUserContext";
 import {
+  CreateWatchlistDataEntry,
   PopulateFromFirestore,
   SaveToFirestore,
 } from "../Components/Firestore";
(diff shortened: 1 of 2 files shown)


src/Hooks/useAuthActions.js
```diff
@@ -91,6 +100,7 @@ export default function useAuthActions() {
       const provider = new GoogleAuthProvider();
       const userCredential = await linkWithPopup(auth, provider);
       await onRegistrationSuccess(userCredential, "google");
+      await writeAllWatchlistDataEntries();
```
metzsg, line 103: Is there a way to merge the watchlist data writing logic into `onRegistrationSuccess`?  I'd like to keep the individual likeWith*/registerWith* functions below free of business logic.  Maybe something after the localuser data is saved like:

```js
// On a non-anonymous registration, write watchlist data for all watchlists.
if (isRegistration && authProvider !== "anonymous") {
  await writeAllWatchlistDataEntries();
}
```

Should be a no-op for first-time registrations, since their `lists` arrays should be empty, and so `await Promise.all([])` will immediately resolve.  Wdyt?

metzsg (approved):

src/Hooks/useAuthActions.js
KennethMetz: Sounds good to me! For the no-op, I added a conditional statement checking whether `localUser.lists.length > 0`.

PR #176, opened 2023-07-20
Fixes /register redirects

KennethMetz:
- Allows anonymous users to stay on `/register` without being auto-forwarded to `/home`.

- Redirects users looking to register an account, but without any 'likes' to `/onboarding` instead of `/home`, for a shorter user journey.

### DIFF
--- a/src/Components/Onboarding.js
+++ b/src/Components/Onboarding.js
@@ -41,7 +41,7 @@ export default function Onboarding() {
     if (!user) return;
     // Let anonymous users register accounts and
     // redirect existing authorized users
-    if (!user?.isAnonymous) return navigate("/home");
+    if (!user.isAnonymous) navigate("/home");
   }, [user, localUser]);
 
   return (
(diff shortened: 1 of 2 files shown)


src/Components/Onboarding.js
```diff
@@ -38,8 +38,10 @@ export default function Onboarding() {
   const navigate = useNavigate();
 
   useEffect(() => {
-    //Check for existing user auth
-    if (user) return navigate("/home");
+    if (!user) return;
+    // Let anonymous users register accounts and
+    // redirect existing authorized users
+    if (!user?.isAnonymous) return navigate("/home");
```
metzsg, line 44: `useEffect` functions return value should be a function that will be invoked as a clean-up the next time the `useEffect` function gets run.  So, here, instead of `return navigate("/home");`, you can probably just do `navigate("/home")` and let the function return undefined implicitly.

Also, nit: if we know user has value here (since there was an early return above if it didn't), then the `?` is not needed.

metzsg (approved):
Looks good! Thanks for the fixes

src/Components/Onboarding.js
KennethMetz: Great comments!!

PR #177, opened 2023-07-24
Add greeting/explainer

KennethMetz:
1. Adds a greeting/explainer blurb to `/home` page for all guest accounts, that includes a registration 'upsell'
1. The greeting/explainer is shown to unregistered accounts every time they navigate to `/home`.  There is an option to close the blub...but it will re-open each time they re-visit `/home`. (This could get annoying, so I'd be happy to re-work this if desired)
1. Changes the user registration flow.  Now `/onboarding` sends everyone to `/home` as a guest: there is no longer a registration prompt.  This is done to reward to the user with content faster, avoid losing users that don't want to register, and make the logic for determining when to show the greeting/explainer simpler.

### DIFF
--- a/src/Components/GreetingExplainer.js
+++ b/src/Components/GreetingExplainer.js
@@ -4,11 +4,11 @@ import useTheme from "@mui/material/styles/useTheme";
 import Grid from "@mui/material/Grid";
 import Typography from "@mui/material/Typography";
 import {
-  Atom,
   Binoculars,
   CaretRight,
   ChatsCircle,
-  Wrench,
+  FlyingSaucer,
+  PuzzlePiece,
   X,
 } from "phosphor-react";
 import { useState } from "react";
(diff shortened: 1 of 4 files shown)


metzsg (approved):
This is really cool!  I like the change to the onboarding flow.  I think we may have fewer users create non-anonymous accounts just since it is more hidden away now that it is isn't on the onboarding path, but maybe more users will actually get to use the site and enjoy it, since there isn't a registration page to bounce off of.  Less friction to get to cool stuff is good.

Regarding the copy in the explainer box, I like what you have a lot and really like the brevity and focus of things.  I messed around with something I was envisioning that was more tutorial-like: https://drive.google.com/file/d/1xssEZghJgvBy7G76wZny1ZaTvby9WtnI/view?usp=sharing.  Lmk what you think.  I can commit this if you like it and we can iterate on it.  Won't be offended if you prefer it how you had it.  I was just wondering if we could use it to give a quick set of steps for how the site works.  I'm not 100% on all the copy yet, though.

KennethMetz:
Good stuff!! Agreed on the trade-offs of this onboarding flow - I'm thinking we can always switch to the old 'retention focused' flow once Edward goes viral!

And I like that tutorial copy a lot!! It really helps to fill in some knowledge gaps about how Ed works right away - so please commit!

I'm thinking it could be helpful UX to add a delayed pop-up reminding guest accounts their stuff will be lost if they don't register - especially if they're spending like 10+ minutes on the site! I'll add to my to do list.

metzsg:
Ok, just pushed a commit with new copy and icons, ptal!

PR #178, opened 2023-07-24
Notifications - Eliminate duplicate keys

KennethMetz:
- Prevents notifications from having duplicate keys (it was unlikely but still possible)

- Increases quantity of notifications retrieved on each 'see more' click from (4) to (8), for a better feel.


metzsg (approved):
Nice!  Good catch on the keys, it definitely seems possible that users could mash the applaud and heart reactions in the same second.

I like increasing the page size, as well.  Loading 8 at a time is still very reasonable.

PR #179, opened 2023-07-24
Saves Emoji Count on Edited Reviews

KennethMetz:
Maintains the emoji count for reviews/comments when they are edited.  

Losing all your reactions felt like too big of a potentially unforeseen consequence for users just wanting to edit a post.


metzsg (approved):
LGTM!

PR #180, opened 2023-07-25
Add info text when reaction buttons are disabled

metzsg:
In this case, I changed the tooltip to let the user know why these buttons are disabled and how to enable them.  Lmk if you have thoughts about copy.

I also sunk the Tooltip into the EmojiReactionChip component since the logic for whether the button is enabled lives in there, so the tooltip text determination should probably be made in the same place.


KennethMetz:
Very nice!  I like this a lot...and it is another great way to nudge unregistered accounts to sign up for the full experienc!!

Also, I added a similar change to the 'add a comment/review' button tooltip.

KennethMetz (approved):

PR #182, opened 2023-07-27
Onboarding hover & header tweaks

KennethMetz:
- Deletes unnecessary gap at bottom of `onboarding`
- Deletes duplicated tabIndex on 'login' button
- Adds a slight color to `OnboardingAnimeCard` on :hover
- Resizes the <Link> that bounds the `EdwardMLLogo` so only the logo is clickable


metzsg (approved):
Looks good! I really like the hover effect

PR #183, opened 2023-07-27
Fix: Duplicate guest registrations

KennethMetz:
Deletes a redundant call to 'authActions.registerAnonymously' when clicking 'let's go' button on `/onboarding`.  This was creating a 2nd firestore user auth whenever a new visitor went through the onboarding process.


metzsg (approved):
LGTM!

PR #185, opened 2023-07-30
Add pre-home API pings on onboarding, login

metzsg:
The first load of /home can take a while.  Probably too long.  This happens because the API service uses autoscaling and when no requests have come in for a while, it shuts down all the instances of it.  This saves us $, but has a downside.  When a new API request comes in, a new instance needs to start ("cold load") and this takes a few seconds.  My newly optimized API implementation should cold load faster (like 3s instead of 6s), but this is still slow, when you add actual processing time (~100ms) and network latency on top of that.

One way to get /home to load faster the first time is to wake the API up when the user is on /login or /onboarding, since they will probably get to /home next.  This gives the instances time to start up so /home loads don't cold-load so many instances.

There is still the problem of the fact that /home makes like 13 API requests (1 /recommend, 5 /anime for shelves, 6 more /anime for random items, plus a /get for the user shelf contents), so these really end up being split between multiple instances.  So even if I pre-warm an instance before /home, odds are some of these might need another new instance that will need to cold-load.  In this commit, I have lifted up the /recommend and shelf fetches so they run before other api requests and have better chance of being handled by the running instances.  The random shelf can take longer to load, since it is at the bottom of the page.

This problem should go away once we have enough traffic that the API never goes more than a few seconds without requests, but until then, this strategy might help smooth load times.  Really the /recommend load time is the most important one, since that is what the user sees first.  Sadly, that is a POST request so it has a pre-flight stage, meaning other requests usually end up beating it to the live instance, and it has a higher-than-usual chance to need a cold load.

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -1,5 +1,6 @@
 import { useQuery } from "@tanstack/react-query";
 
+const fiveSecondsMs = 1000 * 5;
 const fiveMinutesMs = 1000 * 60 * 5;
 
 const apiUrl = `https://api-jet-lfoguxrv7q-uw.a.run.app`;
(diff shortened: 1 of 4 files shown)


KennethMetz:
Good thinking!! Also, pretty impressive work cutting the load time IN HALF!

PR #186, opened 2023-07-31
Make night mode the default setting

KennethMetz:


metzsg (approved):
It is better in every way.

PR #188, opened 2023-08-05
Add 'Reviewers Often Say' tags on DetailedView

metzsg:
In the spirit of not letting perfect be the enemy of good, this is an initial version of the derived quality tags on anime pages.  I think it is actually pretty cool.  I show up to 4 of the most-commonly described qualities, with a 'see more' button to show the rest.  Future improvements might involve either grouping quality tags by sentiment (positive/negative), including more mixed and negative reviews in the review clustering and labeling to derive more negative tags, and ranking qualities by their uniqueness to make sure interesting ones are surfaced before the 'See more', and generic things like "Strong character development" get moved toward the end of the list.

### DIFF
--- a/src/Components/ExpandableTags.js
+++ b/src/Components/ExpandableTags.js
@@ -1,19 +1,18 @@
-import { Button } from "@mui/material/";
 import Box from "@mui/material/Box";
 import Link from "@mui/material/Link";
 import useTheme from "@mui/material/styles/useTheme";
 import { useState } from "react";
 
-export default function ExpandableTags({ items, compactItems }) {
-  compactItems = compactItems ?? 4;
+const compactSize = 4;
 
+export default function ExpandableTags({ items }) {
   const theme = useTheme();
 
   const [expanded, setExpanded] = useState(false);
 
-  const canShorten = items?.length > compactItems;
+  const canShorten = items?.length > compactSize;
 
-  const shownItems = expanded ? items : items?.slice(0, compactItems);
+  const shownItems = expanded ? items : items?.slice(0, compactSize);
 
   return (
     <Box sx={{ display: "flex", flexWrap: "wrap" }}>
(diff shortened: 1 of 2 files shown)


src/Components/ExpandableTags.js
```diff
@@ -0,0 +1,57 @@
+import { Button } from "@mui/material/";
+import Box from "@mui/material/Box";
+import Link from "@mui/material/Link";
+import useTheme from "@mui/material/styles/useTheme";
+import { useState } from "react";
+
+export default function ExpandableTags({ items, compactItems }) {
```
KennethMetz, line 7: Nit: It doesn't look like the `compactItems` parameter is used here, can we delete it?  Or will it be used elsewhere?

src/Components/ExpandableTags.js
```diff
@@ -0,0 +1,57 @@
...
+        <Link
+          color="inherit"
+          variant="body1"
+          underline="none"
+          onClick={(e) => {
+            setExpanded(!expanded);
+            e.stopPropagation();
+          }}
+          component="button"
+          sx={{
```
KennethMetz, line 45: Let's add a `mb:1.5` to the 'see more' button so it's vertically aligned with our tags.

KennethMetz (approved):
I really like this new shelf!  It's great for scanning shows quickly to determine if you want to keep researching it or if you should move on.  Excellent PR!

I added a couple minor comments - no pressure to incorporate either.

src/Components/ExpandableTags.js
metzsg: Good point! I've removed it from the interface and made it a const in the file.

src/Components/ExpandableTags.js
metzsg: Great suggestion, done.

PR #189, opened 2023-08-06
Move watch links to left column on desktop

metzsg:
This gets rid of the big waste of space that usually happens in the "Watch" section on DetailedView on desktop.  I left it where it is on mobile, though, since I quite like our DetailedView header as-is on mobile now.

Also tweaked the watch link icons to make them slightly smaller, and shifted the padding so the left-most watch link aligns properly with the left edge of its content space.

Check it out and lmk what you think.

### DIFF
--- a/src/Components/DetailedView.js
+++ b/src/Components/DetailedView.js
@@ -196,6 +196,16 @@ export default function DetailedView() {
               />
             ))}
           </Box>
+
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
I like it a lot!! Great changes!!!

PR #190, opened 2023-08-07
Emoji Data Restructuring

KennethMetz:
Reworks where emoji data is stored to allow for sorting of comments/reviews by popularity.
- Whether a user has selected an emoji is stored within `users/:uid/reactions/:uniqueEntityId`
- The emoji tally for an item is stored within either `animeData` or `watchlistData` collection, in the same doc as the review/comment information.

Quirks to be aware of:
- On clicking an emoji, we only write to firestore. The visual update of the reaction count on the chip is faked, to eliminate a delay between clicking the chip and seeing the value increment.  Obviously this has serious flaws (ie. the firestore write fails). Will revisit.
- To update the emoji tally I'm using firestore's increment function which can only be called [once per second](https://firebase.blog/posts/2019/03/increment-server-side-cloud-firestore/). I will update to using a [distributed counter](https://firebase.google.com/docs/firestore/solutions/counters) on a future PR (this one's bursting at the seams already!)

Next steps for Ken after we merge:
- Write firestore rules.
- Move data for existing lists/comments/reviews to match new data structure.

### DIFF
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -345,6 +345,9 @@ export async function CreateWatchlistDataEntry(userId, listId) {
     userId: userId,
     listId: listId,
     random: generateRandomWatchlistInt(),
+    applauseCount: 0,
+    heartCount: 0,
+    trashCount: 0,
   };
   const docName = userId + listId;
   let docRef = doc(db, "watchlistData", docName);
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -208,54 +208,44 @@ export function generateId() {
   return doc(collection(db, "test")).id;
 }
 
-export async function getReviewReactions(uid, uniqueEntityId, setReviewRxns) {
-  try {
-    let reactionsRef = doc(db, "users", uid, "reactions", uniqueEntityId);
-    let reactionSnap = await getDoc(reactionsRef);
-    // Use default value if the document doesn't exist yet.
-    if (!reactionSnap.data()) {
-      setReviewRxns({
-        uniqueEntityId: uniqueEntityId,
-        applause: false,
-        heart: false,
-        trash: false,
-      });
-    } else {
-      let temp = reactionSnap.data();
-      setReviewRxns(temp);
-    }
-  } catch (error) {
-    console.error("Error loading data from Firebase Database", error);
+export async function getReviewReactions(uid, uniqueEntityId) {
+  let reactionsRef = doc(db, "users", uid, "reactions", uniqueEntityId);
+  let reactionSnap = await getDoc(reactionsRef);
+  // Use default value if the document doesn't exist yet.
+  if (!reactionSnap.data()) {
+    return {
+      uniqueEntityId: uniqueEntityId,
+      applause: false,
+      heart: false,
+      trash: false,
+    };
+  } else {
+    return reactionSnap.data();
   }
 }
 
 // Handle "watchlistData / reactions" on firestore
 
-export async function getListReactions(uid, docId, setListRxns, setRxnCount) {
-  try {
-    let reactionsRef = doc(db, "users", uid, "reactions", docId);
-    let rxnCountRef = doc(db, "watchlistData", docId);
-    let [reactionSnap, rxnCountSnap] = await Promise.all([
-      getDoc(reactionsRef),
-      getDoc(rxnCountRef),
-    ]);
-    if (!reactionSnap.data()) {
-      // Use default value if the document doesn't exist yet.
-      setListRxns({
-        uniqueEntityId: docId,
-        applause: false,
-        heart: false,
-        trash: false,
-      });
-    } else {
-      let temp = reactionSnap.data();
-      setListRxns(temp);
-    }
-    let temp = rxnCountSnap.data();
-    setRxnCount([temp.applauseCount, temp.heartCount, temp.trashCount]);
-  } catch (error) {
-    console.error("Error loading data from Firebase Database", error);
+export async function getUserReactions(uid, docId) {
+  let reactionsRef = doc(db, "users", uid, "reactions", docId);
+  let reactionSnap = await getDoc(reactionsRef);
+  if (!reactionSnap.data()) {
+    // Use default value if the document doesn't exist yet.
+    return {
+      uniqueEntityId: docId,
+      applause: false,
+      heart: false,
+      trash: false,
+    };
+  } else {
+    return reactionSnap.data();
   }
+}
+
+export async function getReactionCount(uid, docId) {
+  let rxnCountRef = doc(db, "watchlistData", docId);
+  let rxnCountSnap = await getDoc(rxnCountRef);
+  return rxnCountSnap.data();
 }
 
 export async function SaveReactionsToFirestore(
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -213,7 +213,6 @@ export function generateId() {
 }
 
 export async function getReviewReactions(uid, docId, setReviewRxns, type) {
-  console.log(uid);
   try {
     let reactionsRef = doc(db, "users", uid, "reactions", docId);
     let reactionSnap = await getDoc(reactionsRef);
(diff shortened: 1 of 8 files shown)


metzsg (commented):
I really like this update -- moves the data storage structure to something much more scalable and locked-down.  Nice!  It is a big change and I have been working through reading it.  Not quite done yet, but I've left some first-pass comments here for you to check out.  Lmk what you think!

src/Components/Firestore.js
```diff
@@ -211,35 +208,115 @@ export function generateId() {
   return doc(collection(db, "test")).id;
 }
 
+export async function getReviewReactions(uid, uniqueEntityId, setReviewRxns) {
```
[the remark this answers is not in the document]
metzsg: Then a usage in Review.js could be 
```js
  useEffect(() => {
    getReviewReactionsNew(user.uid, uniqueEntityId)
      .then((value) => setItem(value))
      .catch(() => {
        // Show error UI, maybe.
      });
  }, [user.uid, uniqueEntityId]);
```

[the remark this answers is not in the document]
KennethMetz: Yeah, that's a fair point - it does seem to make things messier. I'll revise accordingly!

src/Components/Review.js
```diff
@@ -45,7 +43,8 @@ export default function Review({
   const theme = useTheme();
   const confirm = useConfirm();
 
-  const { data: reviewerInfo } = useProfile(item.uid);
+  const [item, setItem] = useState({});
```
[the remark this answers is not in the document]
KennethMetz: Lol, agreed! This definitely tripped me up during my re-work of it.  

Yes - it should only store the booleans for the users reactions.  I'll definitely incorporate this - future Ken will thank me, I'm sure!

src/Components/ProfileListPage.js
```diff
@@ -69,10 +69,11 @@ export default function ProfileListPage() {
   let listHasDesc = false;
   let showSuggestions = false;
   let privateList = false;
+  const [rxnCount, setRxnCount] = useState([]);
 
   useEffect(() => {
-    GetListReactions(`${userId}${listId}`, setListRxns);
-  }, [listId]);
+    getListReactions(user.uid, `${userId}${listId}`, setListRxns, setRxnCount);
```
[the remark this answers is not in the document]
KennethMetz: Kind of - I haven't moved the firestore data for any of the existing watchlists/reviews yet, so this behavior is expected for watchlists/reviews created prior to this PR.

I was waiting until we were ready to merge to edit any existing data.

For a demonstration of a working watchlist checkout out @TwitterTestAccount's list 'Shonen Jump FTW'.

src/Components/Firestore.js
```diff
@@ -211,35 +208,115 @@ export function generateId() {
...
+        trash: false,
       });
-      return;
+    } else {
+      let temp = reactionSnap.data();
+      setListRxns(temp);
     }
-    let temp = querySnapshot.data();
-    setListRxns(temp);
+    let temp = rxnCountSnap.data();
```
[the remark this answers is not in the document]
KennethMetz: We'll assume this is an invariant going forward (all watchlists/reviews/comments will have that field created during their initial save).

The downside of this is what you're finding...all pre-existing watchlists/reviews/comments won't show valid counts until I update the data manually.

src/Components/Firestore.js
```diff
@@ -211,35 +208,105 @@ export function generateId() {
...
+  }
+}
+
 // Handle "watchlistData / reactions" on firestore
 
-export async function GetListReactions(docId, setListRxns) {
-  try {
-    let animeRef = doc(db, "watchlistData", docId, "reactions", "emojis");
-    let querySnapshot = await getDoc(animeRef);
+export async function getUserReactions(uid, docId) {
```
metzsg, line 229: Is this function the same as `getReviewReactions` above?  If so, do we need both, or should we consolidate to just one?  If so, let's use the generally-named `getUserReactions`

src/Components/Firestore.js
```diff
@@ -211,35 +208,105 @@ export function generateId() {
...
+      trash: false,
+    };
+  } else {
+    return reactionSnap.data();
   }
 }
 
-export async function SaveListReactionsToFirestore(docId, updatedRxns) {
-  let Ref = doc(db, "watchlistData", docId, "reactions", "emojis");
+export async function getReactionCount(uid, docId) {
```
metzsg, line 245: This function only gets reaction counts for watchlists, not for reviews or comments -- should that be clear in the name?

src/Components/Firestore.js
```diff
@@ -211,35 +208,105 @@ export function generateId() {
...
+  updatedRxns,
+  reactionTo,
+  reaction,
+  commentOwnerId
+) {
+  const value = updatedRxns[reaction] ? 1 : -1;
+  let key;
+  if (reaction === "trash") key = "trashCount";
+  else if (reaction === "heart") key = "heartCount";
+  else if (reaction === "applause") key = "applauseCount";
```
metzsg, line 263: How about adding:
```js
else throw new Error(`Unknown reaction: ${reaction}`);
```
so we don't proceed if the reaction type is bad

src/Components/Firestore.js
```diff
@@ -211,35 +208,105 @@ export function generateId() {
...
+  const rxnRef = doc(db, "users", uid, "reactions", uniqueEntityId);
+  let rxnCountRef;
+
+  if (reactionTo === "list") {
+    rxnCountRef = doc(db, "watchlistData", docId);
+  } else if (reactionTo === "reviews") {
+    rxnCountRef = doc(db, "animeData", docId, "reviews", uid);
+  } else if (reactionTo === "comments") {
+    rxnCountRef = doc(db, "watchlistData", docId, "reviews", commentOwnerId);
+  }
```
metzsg, line 277: Similarly, should we throw if `reactionTo` is not known, since we would be proceeding with an undefined `rxnCountRef`, which will surely cause problems?

src/Components/Firestore.js
```diff
@@ -211,35 +208,105 @@ export function generateId() {
...
+// Clears all stored emoji selections for users who reacted to this review/comment
+export async function deleteAllReactions(uniqueEntityId) {
+  const allRxns = query(
+    collectionGroup(db, "reactions"),
+    where("uniqueEntityId", "==", uniqueEntityId)
+  );
+  try {
+    const querySnapshot = await getDocs(allRxns);
+    querySnapshot.forEach((doc) => {
+      deleteDoc(doc.ref);
```
metzsg, line 303: This will require one user (the user doing the deletion) to be able to delete documents from a bunch of other users' `/users/:uid/reactions` collections.  I know we discussed an access-control principal that the owning user is the only one able to write to their `/users/:uid` document and subcollections, and so these deletions would be disallowed if we add such a rule.  

I'm also worried this could result in a lot of deletions all happening from one user's web app, in the case of a review that had a lot of reactions being deleted.  Is there any harm in leaving these zombie reactions in the respective users' `reactions` collections?  Maybe at least util a server-side vacuum job can clean them up?

metzsg (approved):
Thanks for the updates, this is really shaping up nicely!  Left a few final comments, ptal when you have a sec

src/Components/Firestore.js
KennethMetz: Great catch!! Functions have been consolidated.

src/Components/Firestore.js
KennethMetz: All very fair concerns!

The unwanted behavior from zombie reactions, is if someone updates a popular review/comment by deleting it and then reposting (instead of editing it - the way god intended) then all of their friends who had interacted with the old review/comment will still have those emojis marked as selected, despite the emoji count of this new review/comment being 0. 

Admittedly, this is probably an unlikely case, but I did run into it while testing things out so take that for what you will 🙂.  I'm satisfied it we just run a server-side sweeper daily or so.. 

Lmkwyt...but for now I'll delete this `deleteAllReactions` function.

PR #191, opened 2023-08-10
Fixes Notification Popper Bugs

KennethMetz:
- Prevents the `New Notifications` button (within the notifications popper) from flashing on screen when the popper is opened.
- Prevents the notification times from re-rendering during popper's closing animation (bug only really visible while time is in [seconds]).


metzsg (approved):
Looks good!

PR #192, opened 2023-08-10
Fix - Prevent temporary render of incorrect 'For You' score

KennethMetz:
Prevents displaying the 'For You' score based on an empty localUser object for registered users when on a slow network.

- To recreate bug, as a registered user refresh a <detailedView> page with a throttled network.  A preliminary 'For You' score will be displayed (based on an empty localUser) until localUser gets populated from Firestore.


metzsg (approved):
Thanks for the fix! This had been bugging me

PR #193, opened 2023-08-10
Home Page Changes

KennethMetz:
- Allows users without any likes to go directly to /register (forcing them to like a show first doesn't make sense any more - as on /onboarding there's no way to differentiate between folks we want to send to /register next as opposed to the typical flow of hitting /home next)
- Deletes 'register as guest' button from /register. Due to our new flow, users will only land on /register by navigating there in the hope of saving their account. So, having the 'register as guest' button felt unnecessary and possibly even misleading as nothing would get saved.
- Styles the homepage for anonymous users with 0 likes.  I didn't really like how this looked...it wasn't clear the page had any real structure.  Check out the before and after.  Also, feel free to play around with it - I'm tinkering with a few ideas: 
1.  Removing the rainbow or possibly adding a horizontal scrolling effect. 
2. Instead of ghost cards, show some hardcoded cards overlaid with a gradient that makes our Typography legible.

LMKWYT!

**BEFORE**
BeforeSnapshot
 
**AFTER**
AfterSnapshot

### DIFF
--- a/src/Components/Firestore.js
+++ b/src/Components/Firestore.js
@@ -30,7 +30,7 @@ export async function PopulateFromFirestore(user, localUser, setLocalUser) {
     let docRef = doc(db, "users", user.uid);
     let querySnapshot = await getDoc(docRef);
     let data = querySnapshot.data();
-    if (!data.likes) {
+    if (!data?.likes) {
       data = {
         ...data,
         likes: [],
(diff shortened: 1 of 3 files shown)


metzsg (approved):
This one is pretty cool, I like the change on home.  I definitely have thought that the experience for users with no likes needed some improvement, and I like putting the text over cards a lot more.  I'm super interested to continue iterating around with that -- I like some of the suggestions you mentioned and would be curious to see just how aesthetic and interesting we can get!

PR #194, opened 2023-08-14
Fetch random anime via single API call

metzsg:
Since fetching the random items was 6 of the 12 api calls that are made on a /home load, I added support for a 'random' sort via the api's /anime endpoint.  This lets us get a random set of n items with a single api call.  It does return a different set on every call, so I tweaked the react-query settings to get rid of some 'stale' refetches that could happen and make the shelf spontaneously reload without a new page load.  Lmk what you think!

### DIFF
--- a/src/Components/APICalls.js
+++ b/src/Components/APICalls.js
@@ -117,6 +117,24 @@ export function useAnimeTN(selectedGenre) {
   );
 }
 
+export function useAnimeRND() {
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
I like this a lot!!! Avoiding some of those 'stale' refreshes is a great UX...I'm going to look into incorporating that same idea into our community list shelf as well.

PR #195, opened 2023-08-14
Fetch community list via TanStackQuery

KennethMetz:
Envelopes call for a random community list in TanStackQuery, to minimize unnecessary firestore calls and mirror the updated behavior of 'random' shelf as implemented in PR #194


metzsg (approved):
LGTM!

PR #197, opened 2023-08-22
Cleaner breathing edwardML logo

KennethMetz:
This PR address two issues with our existing `<BreathingLogo/>`.

1. Uses a cleaner image. Eliminates the blurry outline that was quite visible on the full Page loading screen when using dark mode.

2. The previous component was inherently hard to maintain. It used SVGator, a subscription service 👎. This PR instead uses the `react-spring` package and a simple svg file (that's been "translated" into jsx, via [svg2jsx](https://svg2jsx.com/)). Now we can easily swap out the logo, change its colors, etc. without having to touch the animation being used.

Additionally, I added breathing logos to the google and twitter buttons on `/register` and `/login` for consistency's sake.


metzsg (approved):
Wow, this looks so much better!  Love the cleaner svg, and the much cleaner code, especially in BreathingLogo.js.  Nice!

PR #198, opened 2023-08-23
Bug fix: Prevents tooltip from obstructing emojis on <ProfileListPage>

KennethMetz:
Using the `onExited()` parameter within `<Grow>` causes the popper to not unmount from the DOM when closed.  As a result, the popper's tooltip obstructs the watchlist emojis when closed. 

To replicate bug:
1. On a user profile with > 4 notifications, go to a `<ProfileListPage>`. 
2. Open the `<NotificationDropMenu>`
3. Expand the menu by clicking `see more`.
4. Close `<NotificationDropMenu>`
5. Emojis for the watchlist will be obstructed by the Tooltip from the closed `<NotificationDropMenu>`.

### DIFF
--- a/src/Components/NotificationDropMenu.js
+++ b/src/Components/NotificationDropMenu.js
@@ -153,12 +153,12 @@ export default function NotificationDropMenu() {
         style={{
           zIndex: "4",
           maxWidth: isMobileWidth ? "95vw" : "530px",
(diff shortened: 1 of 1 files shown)


metzsg (approved):
Nice fix! Thanks!

PR #199, opened 2023-08-28
Feature/reactions rework refactor

metzsg:
Ok, I struggled with this one a fair bit, and what I have now _still_ doesn't do everything your version did (it doesn't write notifications yet). I am impressed you got it working as well as you did.

I think this is pretty close to my idea. Basically, my goal was to reduce the complexity of the reaction system through 1) abstraction, and 2) encapsulation. Basically, to extract all logic for the 'reactions' features out from shared components into their own components, hooks, and util library. As part of that util library, I created a single place where you can see, at-a-glance, where each reaction target writes its data. As a side effect, this refactor makes it really easy to add new reaction targets or types, by changing just one file. There are some downsides, which I will elaborate on. But first, a summary:

Let's start at `ReactionUtil.js`. This file:
- defines methods to get `RxnTarget`s, which are objects that hold all necessary info about the target of a reaction.
- defines an array of `RxnType` objects, each one of which describes a type of reaction that can be made ("love", "trash", etc).
- defines a function to get a default `UserRxns`, an object holding the user's reaction state to a target.
- defines a function to get a default `CountDoc`, an object holding a target's reaction counts. Generally, this file has lots of other stuff in it (a review, a watchlistData doc), so this is just used when we don't find one of those.

Then, I have taken all the code to obtain and update the user's reaction state and moved it to a hook `useReactionState`.

All reaction buttons are wrapped in a single component `EmojiReactionChips`, which creates an `EmojiReactionChipNew` for each `RxnType` defined in `ReactionUtil`. Inside `EmojiReactionChipNew` is some controller logic per-reaction-type. This includes the fake count increment/decrement.

Finally, I extracted all the actual UI-building code to `EmojiReactionChipView`. Probably didn't have to do that, but it helped my thought process to separate out business logic (how it works) from view logic (how it looks).

Finally, I made changes to `Review` and `ProfileListPage` to demonstrate how to use the new `EmojiReactionChips` component, which requires only two things:
- A `RxnTarget` object, obtained from functions in `ReactionUtil`, and
- A CountDoc, the document that has the reaction counts at the time of page load.

Here are the (possible) bad things to consider:
- Dependency data is passed between components as a `RxnTarget` object. This should always be memoized via `useMemo`. If you use a hook that accesses fields in this object, and put the object itself in the deps list of the hook, re-creating the `RxnTarget` object on re-render will count as a 'change' of the dependency, even though all the values in the objects might be the same. So, you gotta remember to memoize these objects to avoid unintended hook executions.
- Some reaction button business logic in `useReactionState` hook, and some in `EmojiReactionChipNew` -- I don't like how the state management is split there, would be cool to unify it.
- Doesn't write notifications yet. I can add this without too much trouble if you approve of this approach.
- Maybe over engineered? Some indirection. Maybe not more than before, but there still is a fair bit. This is a tricky feature to write.

I also haven't done much clean-up in this commit, but we will be able to delete some stuff because of this refactor, if we like it and want to move forward. Lmk what you think! (Start at `ReactionUtil` while reading though)


src/Util/ReactionUtil.js
```diff
@@ -0,0 +1,148 @@
...
+  listOwnerId,
+  commentAuthorId
+) {
+  return {
+    type: "listComment",
+    noun: "comment",
+    entityId: `listComment-${listOwnerId}${listId}-${commentAuthorId}`,
+    countDocPath: [
+      "watchlistData",
+      listOwnerId + listId,
```
KennethMetz, line 83: The `listId` variable already includes the `listOwnerId`, so we'll need to delete `listOwnerId` here.

KennethMetz (approved):
Great PR! I really like how the `ReactionUtil` function really cleans up the emoji components by  abstracting all the weird field path concatenation and tooltip verbage manipulation. Great stuff!!

I think we should proceed with this approach. I left one minor comment on the use of the `reactionCounts`, I don't think it's a showstopper but lmk if you see something I don't!

src/Util/ReactionUtil.js
metzsg: Oh, nice catch!  I think we need a name for the id that is formed when we smush together the list owner id with the list id -- maybe `userListId` or something like that

src/Components/Firestore.js
```diff
@@ -432,3 +434,44 @@ export async function deleteWatchlistDataEntry(userId, listId) {
...
+export async function SaveUserRxStateToFirestore(
+  userId,
+  rxnTarget,
+  rxnType,
+  userRxns,
+  incrementBy
+) {
+  const docRef = doc(db, "users", userId, "reactions", rxnTarget.entityId);
+  const countDocRef = doc(db, rxnTarget.countDocPath);
+  const countFieldPath = `reactionCounts.${rxnType.key}`;
```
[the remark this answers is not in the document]
metzsg: I _think_ we should be able to order by fields in a map, if we use dot notation to reference the field name, similar to what I have to do here to update one field of a map without wiping the others.  Something like `.orderBy('reactionCounts.love', descending: true)`.  This is just based on [this answer](https://stackoverflow.com/questions/70047080/flutter-firestore-how-i-can-order-firestore-query-by-field-value-inside-map), though, I haven't tried it to be 100%.  

If that works, the downside of this approach is still all the dot notation stuff, with the win of organization in the document and making it possible to build a 'default' reactionCounts object .  I'm open to putting the counts root-level -- if so, I'll probably need to add another field to a `RxnTarget` for the `countKey` so they can have a key name like `applauseCount`, similar to how you had it, since a root-level field of `love` or `trash` might be ambiguous or maybe even collide with something.  Let me know which way you prefer.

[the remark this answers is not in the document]
KennethMetz: Good find! It looks like it is possible - who knew!?! 

Lets proceed with the `reactionCount` structure as you have it. You can disregard my original comment.

src/Util/ReactionUtil.js
metzsg: I've actually chosen to keep this function signature requiring both `listId` and `listOwnerId` separately, since I ended up needing them separate in the notification.  I've just updated the call to this method in `Review` to pass in a `listId`, which I piped through from `ReviewContainer`.  Lmk what you think

src/Util/ReactionUtil.js
```diff
@@ -0,0 +1,148 @@
...
+ * @property {string} countDocPath A path to the document with reaction counts.
+ * @property {string} ownerId The owner of this target.  They may be notified.
+ */
+
+/**
+ * Build a `RxnTarget` for an anime review.
+ * @param {number} animeId The id of the anime being reviewed.
+ * @param {string} reviewAuthorId The uid of the author of the reivew.
+ * @returns {RxnTarget} A `RxnTarget` for the specified anime review.
+ */
```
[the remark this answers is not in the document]
metzsg: Thanks! JSDoc is nice because it also gives hints in vscode, etc.  (Still not as good as typescript, since it won't alert you if you are putting the wrong type in a field or returning the wrong kind of value from a fn!)

src/Components/Firestore.js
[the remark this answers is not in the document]
metzsg: Awesome, sounds good! I just pushed some commits to address the issues you found, and add notification support for reactions this way.  Take a look and lmk what you think

src/Hooks/useReactionState.js
```diff
@@ -53,9 +55,42 @@ export default function useReactionState(rxnTarget) {
...
+ * @param {import("../Util/ReactionUtil").RxnType} rxnType
+ * @returns {import("../Util/NotificationUtil").Notification} A notification
+ *  object that can be saved to storage.
+ */
+function getNotification(user, rxnTarget, rxnType) {
+  // TODO Extract basic notification template logic to shared lib.
+  return {
+    interactorId: user.uid, // Person sending the emoji
+    action: rxnType.type,
+    docId: rxnTarget.notifData.docId,
```
KennethMetz, line 87: The key name of "docId" here bothers me a little, it's really the "entityId" now (which was needed to fix the `deleteNotification` bug you caught). But for the sake of not breaking all existing notifications I'm fine with leaving it as is.

KennethMetz (approved):
Looks good to me!! 

I agree, it does appear that was a bug with the `DeleteNotification` function. Great catch!

Also, I understand what you mean by the `notifies` name...but it's good enough imo!

src/Hooks/useReactionState.js
metzsg: Yeah, I agree about that.  The trouble is that the actual document ids for comments are the author's uid, which means these ids are not globally unique (since the author can comment on more than one item).  Otherwise, I'd suggest to just put the review document id in there.  We could put a full path in there instead, `watchlistData/${listOwnerId}${listId}` for watchlists and `watchlistData/${listOwnerId}${listId}/reviews/${reviewAuthorId}` for watchlist reviews.  This would make it kinda docId-esque, and these paths have to be globally unique since they are firebase paths.  

Alternatively, we could either:
1) do a migration later and start writing this field in parallel as both `docId` and `entityId` then eventually remove the `docId` after running a script to update any existing notifications without an `entityId` field,
2) Do (1) above, but with `docPath` as the new field to migrate to, and use paths instead of those generated entityId strings,
3) Do either (1) or (2), but quickly delete all notifications and start over, allowing us to make breaking changes to the notification schema, instead of a migration, or
4) leave it as it is and it'll just be a nice piece of tech debt to confuse interns some day

KennethMetz: My vote is for #4. 

It seems we have bigger fish to fry at the moment, and this is just a minor inconvenience/naming nit. Functionally, everything's working well, so let's leave this to the interns!

PR #200, opened 2023-09-03
WIP: Implement watchlist import from MAL

metzsg:
Not really a pull request as much as creating a space to collaborate on this branch.

I left some notes as 'TODO' comments in here.  I am definitely not set on this UX or UI or even code structure, please feel free to delete whatever you want and move things around.  Note that the `/malLists/:username` API endpoint can be called with a ?fake=1 querystring, which will avoid actually calling the MAL api and just return fake data, to cut down on test requests during development, where you might want to be importing dozens of times.  You can call `APIGetMalLists` with `fake` arg set to false to ask the API to get real data from MAL, while testing username entry, etc.  In reality, we will probably rate-limit users' real syncs so they can't abuse it.


KennethMetz:
I got the import feature working and looking pretty good. This PR isn't quite ready to be merged, I'm still looking for buy in on the below:
- Are we happy with the button placement and dialog copy?
- Syncing updated mal lists _OVERWRITES_ any previously synced lists.
- Currently there's no mention of the MAL account name that these imported lists came from in the UI. Is that something we want?
- I'm realizing it would be nice to allow folks to select whether these imported lists are public/private.  Will add soon.

KennethMetz:
After looking into it, allowing imports to be switched between public/private is a bigger lift than anticipated (it requires adding the ability to change the privacy setting of existing lists).  

As the mal lists had to be public in the first place to get imported, I'm fine with forcing all synced lists to be public.

src/Util/ListUtil.js
```diff
@@ -17,3 +18,89 @@ export function slugifyListName(name) {
...
+          source: getMalListSource(malList),
+          syncDate,
+        },
+      },
+    ];
+  }
+
+  // Only registered users have their watchlistData created and displayed on '/home'
+  if (!privateList && !user.isAnonymous)
+    CreateWatchlistDataEntry(user.uid, listId);
```
metzsg, line 78: Do we want to make this an async function and await completion of the firestore operation here?

src/Components/ProfileSidebar.js
```diff
@@ -36,6 +37,9 @@ export default function ProfileSidebar({ hideDetails }) {
               </Button>
             </div>
           )}
+          {/* {isOwnProfile && profile?.authProvider !== "anonymous" && (
```
metzsg, line 40: nit: Delete if we're not going to have the button here?

src/Components/ProfileImportDialog.js
```diff
@@ -0,0 +1,267 @@
...
+
+  const handleClose = () => {
+    setOpen(false);
+    setAccountName("");
+    clearErrors();
+  };
+
+  function handleFormSubmission() {
+    clearErrors("accountName");
+    if (!errors.accountName) {
```
metzsg, line 46: instead of nesting, I often like a "validate, then act" structure for functions.  if we fail validation, we can take an early out (or throw).  maybe something like:
```js
if (errors.accountName) {
  return; // or throw
}

if (loading) {
  return; // or throw
}

// do action...
```

[This video](https://www.youtube.com/watch?v=CFRhGnuXG-4) explains the philosophy nicely 😀

metzsg (commented):
Ok let's goooo!  This is looking really cool.

I've left a few comments, ptal when you have a sec.

Also, regarding storing the username we synced from, I think that would be a good idea, could be cool to put that info in the watchlist page, maybe "synced from rasengan-rascal on MAL - 3d ago" or something.  Maybe another field in the syncData?

I also don't mind forcing these to be public, since they have to be public on MAL.

src/Components/ProfileImportDialog.js
```diff
@@ -0,0 +1,267 @@
...
+  });
+
+  if (
+    malListsExist?.completed ||
+    malListsExist?.dropped ||
+    malListsExist?.on_hold ||
+    malListsExist?.plan_to_watch ||
+    malListsExist?.watching
+  ) {
+    setLocalUser(temp);
```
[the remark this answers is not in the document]
KennethMetz: Good point!  This was a carry over from a previous iteration of the function...it can be deleted now!

src/Util/ListUtil.js
KennethMetz: Yes!

src/Components/ProfileImportDialog.js
KennethMetz: Great comment. In hindsight, `handleFormSubmission` will only fire if the form submittal passes validation.  

So, we're able to eliminate all of the conditionals I originally had. 🎉🎉🎉

src/Util/ListUtil.js
```diff
@@ -18,4 +23,82 @@ export function slugifyListName(name) {
...
+ * @param {string} desc The users' description of the list.
+ */
+export function createNewWatchlist(
+  name,
+  anime,
+  privateList,
+  desc,
+  user,
+  localUser,
+  malList
```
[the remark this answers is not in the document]
KennethMetz: Fair point! There's definitely some ambiguity there. 😬

I've broken it into two separate functions to help keep `ProfileImportDialog` as small as possible.

KennethMetz (approved):
Comments have been incorporated.

I'm not totally happy with the UI on either `<ProfileListPage>` (it seems like too much text now), or `<WatchlistTile>` (the mal icon looks bad in night mode).

Feel free to tinker with the layout and I'll keep thinking on other options.

PR #201, opened 2023-09-04
Make the 'enter' key submit the email login form

metzsg:
This is just something that has been bugging me.  I like to type my password, then hit enter, instead of clicking a button.  To make this work, I've wrapped the email login form in a `<form>` tag with an `onSubmit` function, and switched the email login button to `type="submit"`.

I also changed the input type for the email input to `type="email"`, which I think will show a special keyboard with an @ sign on mobile, etc.

### DIFF
--- a/src/Components/Login.js
+++ b/src/Components/Login.js
@@ -86,6 +86,11 @@ export default function Login() {
     }
   };
 
+  const handleEmailSubmit = (e) => {
+    e.preventDefault();
+    login("email");
+  };
+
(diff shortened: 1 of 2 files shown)


KennethMetz (approved):
Great improvement - I like this UX much better!!

I would like to also do this for `/register` and a couple of other places. I think it may clash with the way I implemented the `react-hook-form` validation, but I'm sure it can be done!  Stay tuned!

PR #202, opened 2023-09-05
Fix bug with profile cloning when Firebase User is lost

metzsg:
Problem description and demo in this video: https://drive.google.com/file/d/1_hbLZgrXuul9i9G0NdV7wfWAGLrxWm1j/view?usp=sharing

Basically, Safari can lose the Firebase User, which triggers code in `RoutingHelper` to register anonymously.  Unfortunately, the old user's LocalUser data is still present, and the registration code saves that, since usually, during registration, the user does have LocalUser data from onboarding that should be saved.

I've added a solution that takes the user to /login when this happens, since they were already logged in, and lost their authentication.

I've also tried to centralize the 'default LocalUser' definition to one spot, `LocalUserContext`.  Note that there are slight different variations of the default used in various places... `DropMenu` sets avatar and bio to `null` and name to `[]`, while `LocalUserContext` uses `""` for avatar and bio, and doesn't initialize a `name` field.  The logout code in `HandleDialog` seems to have a typo where the field handle is not reset, but instead there is `null: null`.  

I've chosen in my implementation to use empty strings, except for handle, and to initialize the name field to empty string everywhere.  Let's discuss this, though -- not sure if you foresee any issues with that.  There is one catch to setting the name field in the default, which is that it would stomp any name field loaded from firestore in `PopulateFromFirestore` in the `if(!data?.likes)` branch -- I've solved this by expanding the default, and _then_ expanding the data loaded from firestore.  So defaults will all be set, but any values in firestore will be preserved.

### DIFF
--- a/src/Components/DropMenu.js
+++ b/src/Components/DropMenu.js
@@ -15,7 +15,7 @@ import useTheme from "@mui/material/styles/useTheme";
 import { Moon, SignOut, Sun, User, UserCirclePlus } from "phosphor-react";
 import { useNavigate } from "react-router-dom";
 import { auth, logout } from "./Firebase";
-import { LocalUserContext } from "./LocalUserContext";
+import { LocalUserContext, getDefaultLocalUser } from "./LocalUserContext";
 import { useAuthState } from "react-firebase-hooks/auth";
 
 import AppSettingsContext from "./AppSettingsContext";
@@ -88,19 +88,7 @@ export default function DropMenu() {
   async function sendToLogout(e) {
     navigate("/logout");
     await logout();
-    setLocalUser({
-      name: [],
-      likes: [],
-      dislikes: [],
-      lists: [],
-      savedLists: [],
-      avatar: null,
-      bio: null,
-      top8: [],
-      reviews: [],
-      comments: [],
-      handle: null,
-    });
+    setLocalUser(getDefaultLocalUser());
     handleClose(e);
   }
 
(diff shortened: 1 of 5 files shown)


KennethMetz (approved):
Great work finding the cause of this weird behavior!!

This all looks good to me and the `name: ""` field works well. Also, I like the centralized 'localUser' creation, much cleaner!

(I added one comment to a chunk of code I almost deleted before remembering why it was there, lol)

metzsg:
> Great work finding the cause of this weird behavior!!
> 
> This all looks good to me and the `name: ""` field works well. Also, I like the centralized 'localUser' creation, much cleaner!
> 
> (I added one comment to a chunk of code I almost deleted before remembering why it was there, lol)

Nice! I actually think we _could_ delete those lines, if we got rid of the `if (!data?.likes) {` around the merging of `...getDefaultLocalUser()` and `...data`.  Since the default comes first, any values in actual data will overwrite default values, but any keys missing in data will be filled in from defaults. What do you think?

KennethMetz:
> > Great work finding the cause of this weird behavior!!
> > This all looks good to me and the `name: ""` field works well. Also, I like the centralized 'localUser' creation, much cleaner!
> > (I added one comment to a chunk of code I almost deleted before remembering why it was there, lol)
> 
> Nice! I actually think we _could_ delete those lines, if we got rid of the `if (!data?.likes) {` around the merging of `...getDefaultLocalUser()` and `...data`. Since the default comes first, any values in actual data will overwrite default values, but any keys missing in data will be filled in from defaults. What do you think?

I like it a lot!! I pushed a commit to delete the unnecessary conditional statements, so it's ready to merge in my book!

PR #203, opened 2023-09-22
Don't refetch recs on window focus.

metzsg:
When the EdwardML browser window/tab is selected, if five minutes have passed since Home page load, react-query will re-fetch the recommendations.  This was pretty much hidden before when the recommendations were deterministic, but now that I am extending that system to involve some randomness, it means sometimes you come back to the browser window and your recs shift around after a second.  It would be better to make the user explicitly request the refetch, by navigation events, reloading the browser, updating their likes, etc.


KennethMetz (approved):
Nice improvement! 💯!!!

PR #205, opened 2023-10-10
Make the 'enter' key submit the email registration form

KennethMetz:
Refactors /register to make this happen. Implements the [Controller component](https://react-hook-form.com/docs/usecontroller/controller) from react-hook-form to allow for better interaction with our controlled components. 

This change incorporates more useful feedback to the user by re-validating form input onChange after an error message appears, instead of waiting for onBlur.

Additionally, cleans up some unused imports.


metzsg (approved):
Looks great!  I like this 'controller' pattern a lot 💥

PR #206, opened 2023-10-16
Loading Indicator on "For You" Refetch

KennethMetz:
When 'For You' recommendations are being refetched, an indeterminate linear progress bar is displayed.

I don't really like how choppy the progress bar looks during fast queries. I've added a slight delay to prevent the worst of the UI flashes - but on my machine a load time of around 1 second still looks unpolished.  It looks great on slow 3G though!!


metzsg (approved):
Nice! This does the job of at least letting the user know things are refetching when they change their likes while on home.  I agree that is gets a little frame-y when things load fast, and I don't love that either.  The delay is a nice trick to keep it from showing up for super short intervals.

PR #207, opened 2023-11-06
Don't refetch DataFromEdward on window focus && Stop DataFromEdward Flash

KennethMetz:
1. Stops refetch of 'DataFromEdward' on window focus.  (An extension of a previous PR: #203)
2. Prevents the display of 'DataFromEdward' if our Google Authorization call is not complete.  We were getting and displaying the 'DataFromEdward' analysis info for the default empty localUser object. This caused a nasty flash and a clunky animation effect.


metzsg (approved):
Nice fix! Looks good

PR #212, opened 2024-05-01
Implement sitemap generator tool, add sitemap

metzsg:
Search engine crawlers can use an optional sitemap (specified in robots.txt) as a guide for what urls to visit. Anime pages in our app are not linked to directly from the landing page, and generally require some clicking around and onboarding to discover.  Since crawlers won't do this stuff, the sitemap will instead provide them with the urls of all anime pages, so these can be discovered and indexed.  It would be cool for our "Akira" page to come up in search results when someone googles "Akira reviews", etc.


tools/build-sitemap.js
```diff
@@ -0,0 +1,60 @@
...
+    `&page=${page}`;
+  const response = await fetch(url);
+  const data = await response.json();
+  return { items: data.items, hasNextPage: data.items.length === pageSize };
+}
+
+async function fetchAllAnimeIds() {
+  const ids = [];
+  let page = 1;
+  while (true) {
```
KennethMetz, line 22: Slick!!!

KennethMetz (approved):
LGTM! Thanks for putting that together!

